PR #32, opened 2015-01-13
Allow building doc from top level dir, add cheatsheet

gdestuynder:
Notes:
- I think rst2html is the default program name, I made it configurable and it defaults to the original value (rst2html5) but this probably only works on Fedora by default (?)
- I would think doc should be auto rebuilt / cleaned with make all/make clean - however, since the html output and others is dependent on a lot of things i didnt add that yet
- consequently my cheatsheet.rst.html document might be slightly different from yours and you might want to run make doc again :/
- /!\ :warning:  the file in doc/.files/createdb.sh seems to be missing from the tree so the doc building doesn't 100% succeeds. I suppose you have it locally @jvehent

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -243,6 +243,9 @@ rpm-api: mig-api
 	fpm -C tmp -n mig-api --license GPL --vendor mozilla --description "Mozilla InvestiGator API" \
 		-m "Mozilla OpSec" --url http://mig.mozilla.org --architecture $(FPMARCH) -v $(BUILDREV) -s dir -t rpm .
 
+doc:
+	make -C doc doc
+
 test: mig-agent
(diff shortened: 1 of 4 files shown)


jvehent:
screenshot
The titles are big for the amount of text. I'd recommend use <h3> for the cheats:

h1: Modules
h2: File
h3: <name of the cheat>

gdestuynder:
My understanding is that RST automatically assigns headers in order - which I used so that the TOC also gets filled. It's difficult to work around that without making it look like a terrible hack I think (?).

I added commit f2f4cf8 which changes headers to a list as a trade-off (TOC won't get filled tho).

jvehent:
Danke Schon!

PR #35, opened 2015-02-23
in standalone_install.sh ensure proxy isn't used for tests

ameihm0912:
Make sure any proxy related environment variables are not present when we run API tests using curl


jvehent:
Sorry for being pedantic, but for the sake of consistency, could you prefix your commit message with `[minor]` ?

ameihm0912:
Closing to resubmit

PR #39, opened 2015-03-01
Update LICENSE

GabrielDuque:


jvehent:
"The current version of the license is MPL 2.0".
https://www.mozilla.org/en-US/MPL/

PR #40, opened 2015-03-10
[major] Implement support for event workers

jvehent:
This patch implement a new rabbitmq exchange called migevent where
the scheduler publishes any event of interest. Workers can subscribe
to specific events using binding keys (standard rabbitmq topic
exchange). A simple mozdef-asset worker is implemented, and the base
for an agent verification worker is also present but not yet functional.


jvehent:
@ameihm0912 - May I ask you for a review?

jvehent:
I'm merging this. Feel free to review post-merge :)

PR #42, opened 2015-03-13
Bug 104714: implement HasIPConnected on Darwin

djmitche:
I tested this by hand on OS X 10.8.

I did run into some trouble, indicated by the comments inline, where netstat produces a truncated ipv6 address.  I don't have any idea how to fix that, or if it's even possible.

### DIFF
--- a/src/mig/modules/netstat/netstat_darwin.go
+++ b/src/mig/modules/netstat/netstat_darwin.go
@@ -66,7 +66,7 @@ func HasSeenMac(val string) (found bool, elements []element, err error) {
 
 var endpoint_re = regexp.MustCompile("^(.*?)(%[a-z0-9]+)?\\.(\\*|[0-9]+)$")
 
-func parseEndpointString(str string) (ip net.IP, port int) {
+func parseEndpointString(str string) (ip net.IP, port int, err error) {
 	// Note that 'netstat' will sometimes truncate a long IPv6 address, in
 	// which case this function may return an incorrect address or (if the
 	// result ends with ':') nil.
@@ -74,7 +74,7 @@ func parseEndpointString(str string) (ip net.IP, port int) {
 	matches := endpoint_re.FindStringSubmatch(str)
 	if matches != nil {
 		if matches[1] == "*" {
-			return nil, -1
+			return nil, -1, nil
 		}
 		ip = net.ParseIP(matches[1])
 		if matches[3] == "*" {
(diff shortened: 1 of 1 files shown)


src/mig/modules/netstat/netstat_darwin.go
```diff
@@ -62,8 +64,104 @@ func HasSeenMac(val string) (found bool, elements []element, err error) {
 	return
 }
 
+func parseEndpointString(str string) (ip net.IP, port int) {
+	re, err := regexp.Compile("^(.*?)(%[a-z0-9]+)?\\.(\\*|[0-9]+)$")
```
djmitche: It'd be nice to not re-compile this regexp every time -- tips?

djmitche:
@jvehent pointed out http://stackoverflow.com/questions/5390164/getting-routing-table-on-macosx-programmatically -- not sure that's worth the effort..

src/mig/modules/netstat/netstat_darwin.go
```diff
@@ -62,8 +64,104 @@ func HasSeenMac(val string) (found bool, elements []element, err error) {
...
+	if err != nil {
+		panic(err)
```
jvehent: I would recommend not panicking directly, but instead return an error value here, or panic and recover.

src/mig/modules/netstat/netstat_darwin.go
```diff
@@ -62,8 +64,104 @@ func HasSeenMac(val string) (found bool, elements []element, err error) {
...
+		if matches[1] == "*" {
+			return nil, -1
+		}
+		ip = net.ParseIP(matches[1])
+		if matches[3] == "*" {
+			port = -1
+		} else {
+			p, err := strconv.Atoi(matches[3])
+			if err != nil {
+				panic(err)
```
jvehent: Same comment about panicking

src/mig/modules/netstat/netstat_darwin.go
jvehent: You could define it as a global variable, since the regexp is static:

```go
package main

import (
	"fmt"
	"regexp"
)

var re = regexp.MustCompile("^$")

func main() {
	if re.MatchString("test") {
		fmt.Println("test doesn't match")
	}
	if re.MatchString("") {
		fmt.Println("empty string matches")
	}
}
```

jvehent:
Looks good. Just a couple minors.

djmitche:
OK, those should be fixed up - let me know what you think

jvehent:
Looks good. r+

PR #46, opened 2015-03-15
[major/bug] prevent socket stat from locking agent results publication

jvehent:
r? @ameihm0912


ameihm0912:
r+
Looks good to me.

PR #51, opened 2015-04-16
Rework module interface to take parameters from STDIN

jvehent:
Complete rework on the module interface, much cleaner, and uses stdin instead of command line parameters for module input.

@ameihm0912 : could you read the doc and tell me what you think? https://github.com/mozilla/mig/blob/8173ad5daa5808e0fcad78fe0bf5be6c3414871c/doc/modules.rst

The `example` module has also been rewritten to use the new interface: https://github.com/mozilla/mig/blob/8173ad5daa5808e0fcad78fe0bf5be6c3414871c/src/mig/modules/example/example.go


doc/modules.rst
```diff
@@ -10,339 +10,711 @@ In this document, we explain how modules are written and integrated into MIG.
...
-checksum, ... Another module is called `connected`, and looks for IP addresses
-currently connected to an endpoint. `user` is a module to manages users, etc...
+checksum, ... Another module is called `netstat`, and looks for IP addresses
+currently connected to an endpoint. `ping` is a module to ping targets from
+endpoints, etc..
 
 Module are somewhat autonomous. They can be developped outside of the MIG code
 base, and only imported during compilation of the agent. Go does not provide a
-way to load external libraries, so modules are shipped within the agent's static
+way to load modules dynamically, so modules compiled into the agent's static
```
ameihm0912: so modules are compiled into

src/mig/modules/agentdestroy/agentdestroy.go
```diff
@@ -22,7 +22,7 @@ import (
 func init() {
 	mig.RegisterModule("agentdestroy", func() interface{} {
 		return new(Runner)
-	})
+	}, false)
```
ameihm0912: I'm guessing these are all still to be converted since RegisterModule is no more?

ameihm0912:
r+
addition of the message classes is cool.
do you think a "response" from the module should be a class as well? this might be an interesting way of opening the door to the module -> module communications as well, but could be overkill right now, but would basically standardize on a single message format for input and output from modules?

jvehent:
Module output is already somewhat standardized in `modules.Result`. The issue with doing module->module is transforming the output of module A into a valid input for module B. I think we could implement it with the structs proposed in this PR and a data transformation layer like the one described in https://bugzilla.mozilla.org/show_bug.cgi?id=1155981

src/mig/modules/agentdestroy/agentdestroy.go
jvehent: That's right. I need to finish converting the other modules, I just implemented the changes in `example` for now as a proof of concept.

jvehent:
Do you think there's value in using the same base struct for input and output?
I don't know if it buys us much. It does add an extra `switch{case}` to detect the type of message being processed.

ameihm0912:
I'm not sure, that's sort of what I was thinking with something like {"class": "results", "results": {}}
It may be unnecessary though (right now at least).

PR #53, opened 2015-04-25
Add tests

djmitche:
@jvehent I left off in the middle of writing tests for ping, but I'd like to get your feedback before I proceed.  I'm particularly interested in a better way to do the dependency injection or otherwise fake out the system-specific stuff.

### DIFF
--- a/doc/module_memory.html
+++ b/doc/module_memory.html
@@ -1,0 +1,91 @@
+<!DOCTYPE html>
+<html>
+<head>
+    <title>Mozilla InvestiGator: Memory module</title>
+    <meta charset="utf-8" />
+    <link href="docstyle.css" rel="stylesheet" />
+    <meta content="Julien Vehent &lt;jvehent@mozilla.com&gt;" name="author" />
+</head>
+<body>
+    <h1>Mozilla InvestiGator: Memory module</h1>
+    <aside class="topic contents" id="table-of-contents">
+        <h1>Table of Contents</h1>
+        <ul class="auto-toc">
+            <li><a href="#usage">1   Usage</a>
+                <ul class="auto-toc">
+                    <li><a href="#filters">1.1   Filters</a></li>
+                    <li><a href="#options">1.2   Options</a></li>
+                </ul>
+            </li>
+            <li><a href="#memory-scanning-algorithm">2   Memory scanning algorithm</a></li>
+        </ul>
+    </aside>
+    <p>The memory module (MM) allows an investigator to inspect the content of the memory of running processes without impacting the stability of a system. MM implements the <a href="https://github.com/mozilla/masche">Masche</a> cross-platform memory scanning library.</p>
+    <section id="usage">
+        <h2>1   Usage</h2>
+        <p>MM implements searches that are defined by a search label. A search can have a number of search parameters and options, defined below. There is no maximum number of searches that can be performed by a single invocation of MM. The module optimize searches to only scan the memory of each process once, and for each buffer scanned, runs all needed checks on it.</p>
+        <p>MM can filter processes on their name or their linked libraries. A standard way to use the module is to set the <cite>MatchAll</cite> option to <cite>true</cite> and specify a name and a content or byte string to search for. MM will first filter the processes that match the name, a cheap check to perform, and because <cite>MatchAll</cite> is true, will only peek inside the memory of selected processes.</p>
+        <p>Without <cite>MatchAll</cite>, all checks are ran on all processes, which can be very costly on a system that has a large memory usage.</p>
+        <p>In JSON format, searches are defined as a json object where each search has a label (key) and search parameters (value).</p>
+        <p>A search label is a string between 1 and 64 characters, composed of letter [a-zA-z], numbers [0-9], underscore [_] or dashes [-].</p>
+        <pre class="code json"><span class="p">{</span>
+    <span class="nt">"searches"</span><span class="p">:</span> <span class="p">{</span>
+        <span class="nt">"somesearchlabel"</span><span class="p">:</span> <span class="p">{</span>
+            <span class="nt">"names"</span><span class="p">:</span> <span class="p">[</span>
+                <span class="s2">"firefox"</span>
+            <span class="p">],</span>
+            <span class="nt">"contents"</span><span class="p">:</span> <span class="p">[</span>
+                <span class="s2">"some bogus content"</span>
+            <span class="p">]</span>
+        <span class="p">},</span>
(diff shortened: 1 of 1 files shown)


jvehent:
First of all: this is awesome! Thanks a lot for the PR!
I'll try to set aside some time to review this in the next few days. Maybe @ameihm0912 will be able to take a look as well?
I think that some patches will conflict with the work in PR #51 , so maybe we want to stage this work for now and merge after the new interfaces are deployed.

djmitche:
I'm happy to rebase as necessary -- I want to build a testing framework you won't want to throw out in a month :)

jvehent:
No risk of that! I definitely want a test framework for MIG. The new interfaces should be done tomorrow, we can take it from there.

jvehent:
This is now unblocked by the merge of #51 . If you could rebase it against master I'll do a review.

djmitche:
Wow, that's a WHOLE lot harder to mock.  Any suggestions on how to provide parameters to a module now?

jvehent:
I'll have a look and try to implement some tests for the memory module I'm currently writing.
Have you looked at the tests in the standard lib, particularly for the `exec` package? There may be some good ideas there: http://golang.org/src/os/exec/exec_test.go

djmitche:
The particularly hard part is that the code now expects to read from an io (a specific io; and for some reason it keeps reading from that io even after EOF?).  I'd like some generic way to write a test that passes in a set of parameters to Run and then examines the result.

I also need a way to do dependency injection so that I can mock out library calls (`LookupHost`, for example, in the example module).

jvehent:
@ameihm0912 fixed the EOF issue, it's being merged in #56 . The interface change in the PR includes an io.Reader now being passed to `Run()`. The agent sets the reader to stdin, but you can set it to whatever buffer or pipe you'd like in your tests.

I don't have a solution for dependency injection. From what I read, people either let the library use the network, or spin up a fake local server to test against (for things like APIs). Do you think that doing a lookuphost against the network is going to be an issue?

src/mig/modules/example/example.go
```diff
@@ -48,6 +52,11 @@ func init() {
 type Runner struct {
 	Parameters params
 	Results    modules.Result
+
+	// dependency injections
+	hostnameFn       func() (string, error)
+	interfaceAddrsFn func() ([]net.Addr, error)
+	lookupHostFn     func(name string) (addrs []string, err error)
```
jvehent: I don't follow the logic. Do you mean to use those as interfaces to the `net` functions?

src/mig/testutil/modtest.go
```diff
@@ -0,0 +1,60 @@
...
+
+func RunModule(t *testing.T, m mig.Moduler, args interface{}) (mig.ModuleResult, []byte) {
+    var marshalled_args []byte
+    marshalled_args, err := json.Marshal(args)
+    if (err != nil) {
+        t.Error(err)
+    }
+
+    t.Logf("Running module with args %s", marshalled_args)
+    marshalled_res := []byte(m.Run(marshalled_args))
```
jvehent: With the new `Run()` interface, you can do something similar but using a `bytes.Buffer` to pass the `marshalled_args`.
```go
buf := bytes.NewBuffer(marshalled_args)
marshalled_res := []byte(m.Run(buf))
```

src/mig/modules/example/example_test.go
```diff
@@ -0,0 +1,104 @@
...
+}
+
+func TestRunGetHostname(t *testing.T) {
+	r := new(Runner)
+	r.hostnameFn = func() (string, error) {
+		return "myhostname", nil
+	}
+	res, marshalled := testutil.RunModule(t, r, params{true, false, ""})
+	testutil.AssertModuleSucceeded(t, res)
+	assert.Equal(t, res.FoundAnything, true)
```
jvehent: is this really better with `assert` than without?
```go
if !res.FoundAnything {
    t.Errorf("should have found something and didn't")
}
```

jvehent:
It looks like a sound approach overall (but I have limited experience with unit testing). I'll implement some tests for the memory module tomorrow and see how it goes.
About `assert`, is it used only as a convenience to save space in the tests, or does it provide other functionalities?

src/mig/modules/example/example.go
djmitche: These are where I inject the dependencies.  So in normal operation, `lookupHostFn` points to `net.LookupHost`, but in tests, it points to a fake.

djmitche:
Calling system-specific interfaces generally makes flaky, fragile tests that are very sensitive to the environment they're run in, and to external changes.  For example, if you look up `example.com` and someday whoever owns that domain gets tired of paying for it or has an outage, your tests will fail and you'll spend time figuring out why.  Also, if your unit test requires spinning up a docker container or interacts with a hosted database, then it's more of an integration test, and not a unit test.  To be unit testable, code should be easily isolated from its surroundings and tested directly.  DI is one way to do that: all of the external dependencies can be overridden with something local for tests.  But that requires support from the system under test.

As for `assert.Equal` and friends, they start to shine when things *aren't* equal, as they'll print out the different values, helping you diagnose the issue.

djmitche:
http://openmymind.net/Dependency-Injection-In-Go/

jvehent:
Thanks, that's a good read. I agree that we shouldn't depend on system-specific functions for tests. I'm having the issues in the memory module where the tests pass on my machine but not when ran by TravisCI. That's not a viable strategy.

The case of `os.Hostname()` is interesting because it's not an interface and can't be reimplemented cleanly in a test. Dependency injection seems the only viable option, but I somewhat dislike how intrusive it is...

I think we could use interfaces for a lot of testing, and rely on DI only when necessary. It seems to be the way recommended by the Go team: https://talks.golang.org/2014/testing.slide#22

I don't know yet how much refactoring work will be needed to write tests for `mig/client`, but most functions are implemented under the `Client` type, so I presume we could write test interfaces. (http://godoc.org/github.com/mozilla/mig/src/mig/client) For other packages, like `mig/pgp`, we'll probably need to rewrite a good amount to make them testable.

Modules are independent packages so I would prefer to start there and not touch the rest of the code for now. Once we have a good strategy for testing modules, we can propagate it to other places. So how about you repurpose this pull request to test the `example` and `ping` modules, and we'll take it from there?

djmitche:
Sounds good.  That was more or less the original intent - test the example and ping modules - so it's not even much of a "repurpose"

djmitche:
Oops, I rebased, so we'll need a new PR

PR #58, opened 2015-05-11
factor out testing for correct module registration

djmitche:
This adds a test script for each existing module, which just verifies that it's registered and that the registered function returns an object that can be cast to the `modules.Moduler` interface.  Baby steps :)


jvehent:
Looks reasonable. r+

PR #59, opened 2015-05-11
rm ulfr-specific testing harness

djmitche:
This file confuses `go test` because it's named `*_test.go`.  Since it contains `/home/ulfr` I'm assuming it's not meant to be generally useful.


jvehent:
it was meant to be included in the godoc, but isn't useful indeed.

PR #60, opened 2015-05-11
Hide the notes about readline-devel

djmitche:
Since readline is the last package go-gotten, it appears to users that
the install has failed, and sends them looking for how to install
readline and why the install isn't working, when in fact the go module
has installed just fine.

This uses a little make trickery to not give any indication that
anything's wrong unless the install fails.


jvehent:
Ha, yes, thank you for this. It's been itching me to find a fix for it for a while :)

PR #62, opened 2015-05-12
Modules doc fix

djmitche:
The module initialization process is pretty convoluted, and the docs didn't quite match reality.  This adjusts the docs to match reality.  I'll loop back in later PRs and make reality a little more sane.


jvehent:
++ for the doc cleanup!

PR #63, opened 2015-05-22
Add an all-but-deps make target to skip the go-get-deps part

djmitche:
This just saves a little time when I want to run 'make all' repeatedly


jvehent:
r+

PR #67, opened 2015-06-02
[minor] Update cipher ordering

gdestuynder:
Should match mozilla/server-side-tls#69


gdestuynder:
changed commit name. here's a secret squirrel. :squirrel:

jvehent:
r+

PR #68, opened 2015-06-10
[minor] parameter tests for pkg module, and some additional validation

ameihm0912:
Add a parameters test for pkg module, and some additional parameter validation


jvehent:
r+

PR #71, opened 2015-07-03
New mig module to execute data fingerprints

ameihm0912:
This is a new module for MIG that locates information from the file system using data fingerprints. As an example, the module can return version information for existing web framework versions, running kernel versions, installed Python packages, etc.

This initial version uses templates that are hardcoded into the module itself. I think it would be useful in the future to support more dynamic fingerprints that can be supplied from the command line side, but this is hard to do to avoid returning sensitive information from the device, so it's something we would need to think about.

I also implemented a simple file finder and content scanner. I initially looked at using functionality from the file module. But, the path walking functions are deeply tied into the module Runner struct and other parts of the file module itself. As an end result it seemed simpler to just make a small path traversal function. The way it's been added here, it's not tied to any MIG module functionality so it's usable by other modules that may need it as well.

### DIFF
--- a/conf/available_modules.go
+++ b/conf/available_modules.go
@@ -12,6 +12,7 @@ import (
 	_ "mig/modules/netstat"
 	_ "mig/modules/ping"
 	_ "mig/modules/pkg"
+	_ "mig/modules/pkgprint"
 	_ "mig/modules/timedrift"
 	//_ "mig/modules/upgrade"
 	//_ "mig/modules/example"
--- a/src/mig/modules/pkgprint/paramscreator.go
+++ b/src/mig/modules/pkgprint/paramscreator.go
@@ -1,0 +1,53 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/.
+//
+// Contributor: Aaron Meihm ameihm@mozilla.com [:alm]
+package pkgprint
+
+import (
+	"flag"
+	"fmt"
+)
+
+func printHelp(isCmd bool) {
+	dash := ""
+	if isCmd {
+		dash = "-"
+	}
+	fmt.Printf(`Query parameters
+----------------
+%stemplate <name>   - Scan using template
+                    ex: template mediawiki
+		    query for specific module supplied template
+`, dash)
+}
+
+func (r Runner) ParamsParser(args []string) (interface{}, error) {
+	var (
+		fs           flag.FlagSet
+		templateName string
+	)
+
+	if len(args) < 1 || args[0] == "" || args[0] == "help" {
+		printHelp(true)
+		return nil, nil
+	}
+
+	fs.Init("pkgprint", flag.ContinueOnError)
+	fs.StringVar(&templateName, "template", "", "see help")
+	err := fs.Parse(args)
+	if err != nil {
+		return nil, err
+	}
+
+	p := newParameters()
+	if templateName != "" {
+		p.TemplateParams.Name = templateName
+		p.TemplateMode = true
+	}
+
+	r.Parameters = *p
+
+	return r.Parameters, r.ValidateParameters()
+}
--- a/src/mig/modules/pkgprint/paramscreator.go
+++ b/src/mig/modules/pkgprint/paramscreator.go
@@ -28,6 +28,14 @@ func printHelp(isCmd bool) {
 %sroot <path>       - Specify search root
                     ex: root /usr/local
 		    default root is /
+
+Available templates:
+
+linuxkernel         - Running Linux kernel information
+linuxmodules        - Loaded Linux modules
+pythonegg           - Python package versions
+django              - Django framework versions
+mediawiki           - MediaWiki framework versions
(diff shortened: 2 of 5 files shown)


jvehent:
Starting the review. Could you add a doc file in `src/pkgprint/doc.rst` that explains what this does and how to use it?

conf/available_modules.go
```diff
@@ -12,6 +12,7 @@ import (
 	_ "mig/modules/netstat"
 	_ "mig/modules/ping"
 	_ "mig/modules/pkg"
+	_ "mig/modules/pkgprint"
```
jvehent: Is the intent to limit this module to packages fingerprints? If not, would calling it `fingerprint` make more sense?

src/mig/modules/pkgprint/paramscreator.go
```diff
@@ -0,0 +1,75 @@
...
+)
+
+func printHelp(isCmd bool) {
+	dash := ""
+	if isCmd {
+		dash = "-"
+	}
+	fmt.Printf(`Query parameters
+----------------
+%stemplate <name>   - Scan using template
```
jvehent: so, templates must be predefined in the module prior to querying them. is that correct?
Would there be a way to provide a template on the command line?

ameihm0912: Yeah that is correct, currently they are hard coded in the agent. That was the intent with the templateMode boolean in the parameters; if this is false a template from the command line would be used instead of a hardcoded one.

We could permit specifying a template via the command line -- in doing that though it opens up the ability to return information from the devices that could be sensitive (password hashes, etc). I'm not sure of a good way to work around this right now, we could do something like blacklist certain files or directories, I'm not sure that would be ideal either.

jvehent:
Could you also add sample input/output to `doc.rst` ?

src/mig/modules/pkgprint/templates.go
```diff
@@ -0,0 +1,91 @@
...
+//
+// Contributor: Aaron Meihm ameihm@mozilla.com [:alm]
+package pkgprint
+
+import (
+	"fmt"
+	"regexp"
+)
+
+type ppTemplate struct {
```
jvehent: If the end goal is to make these loadable from the command line, you may want to export all fields in the `ppTemplate` and `fingerprint` structs.

jvehent: also, what does `pp` in `ppTemplate` stand for? package print?

src/mig/modules/pkgprint/templates.go
```diff
@@ -0,0 +1,91 @@
...
+	name        string
+	fingerprint fingerprint
```
jvehent: could a template have multiple fingerprints?

src/mig/modules/pkgprint/pkgprint.go
```diff
@@ -0,0 +1,347 @@
...
+	}
+	return res, nil
+}
+
+// Given a file, identify any lines in the file that match the supplied
+// regular expression. Will return a slice of string slices. If the
+// supplied regular expression contains substring matches, each string
+// slice will contain the full line matched, and any substrings, otherwise
+// the slice will only contain the line matched.
+func FileContentCheck(path string, regex string) ([][]string, error) {
```
jvehent: Indeed, this does replicate functionality from the `file` module. Have you look into calling that module yet?

src/mig/modules/pkgprint/pkgprint.go
```diff
@@ -0,0 +1,347 @@
...
+		err error
+	)
+	fp := getTemplateFingerprint(tname)
+	if fp == nil {
+		return res, fmt.Errorf("invalid template name specified")
+	}
+	res.Name = tname
+	s := NewSimpleFileLocator()
+	s.maxDepth = depth
+	if root != "/" {
```
jvehent: won't work on windows...

src/mig/modules/pkgprint/templates.go
```diff
@@ -0,0 +1,91 @@
...
+}
+
+type fingerprint struct {
+	filename     string
+	isRegexp     bool
+	pathFilter   string
+	contentMatch string
+	transform    func(string) (string, error)
+	forceRoot    string
+}
```
jvehent: This looks good, but what would extending it to netstat/memory fingerprints look like? Would we add more fields or should we regroup them into a smaller, more general, set of fields?

For example, there could be a `fpType` method that indicates `file`, `netstat` or `memory` and a `fpSignature` that contains a regex. Or is that too generic?

src/mig/modules/pkgprint/templates.go
```diff
@@ -0,0 +1,91 @@
...
+		"PKG-INFO",
+		false,
+		"egg-info",
+		"^Version: (\\S+)",
+		transformNull,
+		"",
+	}},
+}
+
+func getTemplateFingerprint(name string) *fingerprint {
```
jvehent: Why not make `templates` a `map[string]fingerprint` where `string` is the ppTemplate.name field? That way, you could locate a template using a key lookup instead of iterating.

ameihm0912:
Closing this PR for now, changing the way this module works a bit.

PR #73, opened 2015-07-15
scribe engine integration with MIG

ameihm0912:
WIP PR for scribe integration with MIG.
https://github.com/ameihm0912/scribe
https://godoc.org/github.com/ameihm0912/scribe/src/scribe

For some examples of document syntax, the tests under test/ in scribe repo are a good place to start right now.

Module documentation has not yet been added.

This module is unique in that is the first example of calling another MIG module from within the module, so that part should be reviewed.

It's at the point now where some initial review would be good.

### DIFF
--- a/src/mig/modules/scribe/doc.rst
+++ b/src/mig/modules/scribe/doc.rst
@@ -15,45 +15,46 @@ The scribe module is intended to help support:
 * Executing policy checks on systems, for example as part of using MIG for vulnerability management
 * Execute more advanced file content tests involving dependencies
 
-In addition to supporting on-host analysis using scribe documents, the
-module can also be used to query elements of the host system such as the
-package manager database for installed package information.
-
 This document does not discuss the details around writing scribe tests, the
 scribe project documentation should be reviewed for that. This document focuses
 on usage of the scribe module within MIG and provides some examples.
 
 Usage
 -----
-The scribe module supports two execution modes, scribe document analysis
-and package query mode. To use document analysis mode, the `path` option can
-be given to the module along with the path to a scribe document. To use
-package query mode, the `pkgmatch` option is supplied with a regular expression
-specifying which packages to return information for from the agent.
+Document analysis mode can be used by specifying a document to analyze with
+with `path` option. By default, all tests are returned with a result. To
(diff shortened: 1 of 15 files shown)


ameihm0912:
Most issues noted have been addressed

Note: some additional cleanup and changes are needed here. The modules interface had to be extended slightly to export the standard parameters parser.

ameihm0912:
@jvehent all issues should be resolved.

src/mig/modules/scribe/doc.rst
```diff
@@ -0,0 +1,148 @@
...
+scribe project documentation should be reviewed for that. This document focuses
+on usage of the scribe module within MIG and provides some examples.
+
+Usage
+-----
+The scribe module supports two execution modes, scribe document analysis
+and package query mode. To use document analysis mode, the `path` option can
+be given to the module along with the path to a scribe document. To use
+package query mode, the `pkgmatch` option is supplied with a regular expression
+specifying which packages to return information for from the agent.
```
jvehent: Since both use cases are fairly autonomous, would it make sense to make two separate modules? the `scribe` module would focus on parsing and executing a scribe logic, and the `pkg` module would only provide package querying (using scribe under the hood).

src/mig/modules/scribe/doc.rst
```diff
@@ -0,0 +1,148 @@
...
+In document analysis mode, a JSON document is supplied containing a valid
+scribe document.
+
+A scribe document contains a series of tests. Tests are typically comprised
+of two elements - a test source, which details a source of information from
+the host system, and an evaluator which specifies criteria that must match
+for the test to be true.
+
+The following is a simple document example that validates OpenSSL is at least
+version 1.0.1e. If this test fails, it will return true.
```
jvehent: `If this test fails, it will return true.`
I'm bothered by this. If the test says "list everything that has openssl >= 1.0.1e", then the test should succeed when a host has openssl >= 1.0.1e, not the other way around.

ameihm0912:
@jvehent rebased against master, should be good to go. Please review when you have a chance.

jvehent:
r+ \o/

PR #74, opened 2015-07-18
[minor] remove pkg module, this functionality is integrated into scribe

ameihm0912:
The functionality in pkg has been moved into the scribe module, which
provides a simpler way of performing the same functions.


jvehent:
As discussed, instead of completely removing this module, it would be good to rewrite it to use `scribe` so we can simply invoke `mig pkg -name nginx`.

ameihm0912:
Closing this PR as indicated in previous comment, we will resubmit.

PR #80, opened 2015-08-23
[doc] contributing guidelines

jvehent:
@ameihm0912 : r?

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -46,8 +46,9 @@ full of interesting information.
   as explicit as possible. For example, a module that inspects files on the file
   system would be called `file`. A module that lists packages is called `pkg`.
 
-* Do not set [GOMAXPROCS](http://golang.org/pkg/runtime/#GOMAXPROCS) ever. It is
-  already set to an appropriate value elsewhere.
+* In modules, you can set GOMAXPROCS to 1, or you can leave it unset and it will
+  be defined somewhere else. Never set GOMAXPROCS to something greater than 1,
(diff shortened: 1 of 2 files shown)


CONTRIBUTING.md
```diff
@@ -0,0 +1,61 @@
...
+
+* When writing modules, follow the documentation at [Modules
+  writing](http://mig.mozilla.org/doc/modules.rst.html) and check that:
+    - tests have been added that cover the core functionalities of the module
+    - documentation has been written into modules/<modulename>/doc.rst
+  Please don't pick a cute or clever name for your module. Pick a name that is
+  as explicit as possible. For example, a module that inspects files on the file
+  system would be called `file`. A module that lists packages is called `pkg`.
+
+* Do not set [GOMAXPROCS](http://golang.org/pkg/runtime/#GOMAXPROCS) ever. It is
```
ameihm0912: Is this the case right now? It looks like we explicitly set the number of CPUs to CPU/2 in the agent, but does this carry over into the child/module processes? There are a couple modules that set GOMAXPROCS to 1 right now.

ameihm0912:
@jvehent reviewed, all looks good to me and added one comment.

CONTRIBUTING.md
jvehent: What about this wording?
```
In modules, you can set GOMAXPROCS to 1, or you can leave it unset and it will be defined somewhere else.

Never set GOMAXPROCS to something else than 1 because we care about the impact the agent has on its host.
```

ameihm0912: Sounds good to me.

PR #82, opened 2015-08-25
Add instructions on how to directly invoke mig-agent

gdestuynder:
Note: the html file may need regeneration with ulfr's setup for consistency. I generated it with my system and notice subtles diffs in the css styles from my version of docutils :(


jvehent:
r+

PR #83, opened 2015-08-25
[major] reorganize repository structure for `go get mig.ninja/mig`

jvehent:
This is a major change to the structure of the repository and the Makefile,
but a minor change in terms of functionalities. It simply changes the structure
of the directories to be compliant with Go's idiomatic build methods.

I also bought the domain `mig.ninja` and will now be using that as
a standard import path for everything. The magic to serve everything
under mig.ninja/mig will be documented elsewhere.


jvehent:
@ameihm0912 : r?

ameihm0912:
works as advertised for me

I'm guessing we still need the makefile to build configurations

Is the intent the makefile is still run from the subdirectory under src/ in gopath?

For example, we could always do go build mig.ninja/mig/api, which ties the generation of binaries more completely into GOPATH.

jvehent:
Indeed, we don't need to build from `$GOPATH/src/mig.ninja/mig` anymore. We probably don't even need the makefile for a lot of things. For example, one could get a copy of the MIG command line binary by simply running:
```bash
$ go get mig.ninja/mig/client/cmd
$ ./go/bin/cmd help
./go/bin/cmd - Mozilla InvestiGator command line client
```
The binary is named after the folder is it contained into, so maybe we should rename those folders:
* api -> mig-api
* agent -> mig-agent
* scheduler -> mig-scheduler
* client/cmd -> mig-cmd
* client/console -> mig-console

We could also move them into a `apps` folder to differentiate between imported packages and apps.

jvehent:
I'm changing the hierarchy to make go get work nicely, but we still want to keep the Makefile for two things:
* adding the `ldflags` with the build version
* setting the BINDIR to GOOS/GOARCH when cross-compiling (otherwise they'll overwrite each other)

jvehent:
New structure implemented. This should be ready to merge when the build passes.

PR #84, opened 2015-08-27
Support for negative regexes in file module

jvehent:
@ameihm0912: r?

I still don't have a way to list all the files in a directory and make sure they *all* have a given value (for example: no file without gpgsign=1), but this is still a step forward.


jvehent:
This need more thoughts... the current logic only apply to entire file "this file must not have a line that matches this regex". We need the same logic at the line level of the file so we can combine positive and negative matches on a line: "this file must only have lines that match this regex and don't match that regex".

work in progress...

PR #85, opened 2015-08-29
File module improvements

jvehent:
Several file module improvements to make more complex inspections of file systems.

### DIFF
--- a/client/mig/main.go
+++ b/client/mig/main.go
@@ -6,14 +6,17 @@
 package main
 
 import (
+	"bytes"
+	"encoding/json"
 	"flag"
 	"fmt"
-	"mig.ninja/mig"
-	"mig.ninja/mig/client"
-	"mig.ninja/mig/modules"
 	"os"
 	"os/signal"
 	"time"
+
+	"mig.ninja/mig"
+	"mig.ninja/mig/client"
+	"mig.ninja/mig/modules"
 )
 
 // build version
@@ -36,13 +39,14 @@ usage: %s <module> <global options> <module parameters>
 -render <mode>	defines how results should be rendered:
 		* text (default):	results are printed to the console
 		* map:			results are geolocated and a google map is generated
--t <target>	target to launch the action on. The default targets all online
-		agents (idle and offline agents are ignored).
+-t <target>	target to launch the action on. The default targets all online agents.
+		(idle and offline agents are ignored).
 		examples:
 		* linux agents:          -t "queueloc LIKE 'linux.%%'"
 		* agents named *mysql*:  -t "name like '%%mysql%%'"
 		* proxied linux agents:  -t "queueloc LIKE 'linux.%%' AND environment->>'isproxied' = 'true'"
 		* agents operated by IT: -t "tags#>>'{operator}'='IT'"
+		* run locally		 -t local
 -v		verbose output, includes debug information and raw queries
 
 Progress information is sent to stderr, silence it with "2>/dev/null".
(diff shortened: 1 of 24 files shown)


jvehent:
r? ( @ameihm0912 || @gdestuynder ) whoever has time :)

client/mig/main.go
```diff
@@ -36,13 +39,14 @@ usage: %s <module> <global options> <module parameters>
...
--t <target>	target to launch the action on. The default targets all online
-		agents (idle and offline agents are ignored).
+-t <target>	target to launch the action on. The default targets all online agents.
+		(idle and offline agents are ignored).
 		examples:
 		* linux agents:          -t "queueloc LIKE 'linux.%%'"
 		* agents named *mysql*:  -t "name like '%%mysql%%'"
 		* proxied linux agents:  -t "queueloc LIKE 'linux.%%' AND environment->>'isproxied' = 'true'"
 		* agents operated by IT: -t "tags#>>'{operator}'='IT'"
+		* run locally		 -t local
```
gdestuynder: nit "run on local system" (just to avoid any confusion)

modules/file/doc.rst
```diff
@@ -163,8 +163,52 @@ Several options can be applied to a search:
...
+  a file. By default, content regexes are applied at the file level and will
+  return a match if one line matches one regex, and if another line matches another
+  regex. When the `macroal` option is set, each line in the file must match all
+  content regexes defined in a given search to return a match. It is set to not
+  set by default.
+
+  example: `-path /home -name authorized_keys -content "^(#|\s+|...list of ssh keys...)$" -macroal`
+
+  will list authorized_keys file that have contain either a comment, an empty
+  line or one of the listed ssh keys. If a file has a line that doesn't match
```
gdestuynder: nit: clarifications also as this option seem rather special:

If the file has ANY line that doesn't match

It may be useful to insert that this is here as a different way to do a "not" match/search (which Im guessing is the intent), so instead of "not matching x" or doing a "grep -v" you can do a very explicit selection in the file

modules/file/doc.rst
```diff
@@ -163,8 +163,52 @@ Several options can be applied to a search:
...
+  the content regex, it will not be returned in the results.
+
+* **mismatch=<filter>** inverts the results for the given filter. This can be used
```
gdestuynder: Just a comment: this argument seem particularly complex to comprehend without reading this and the macroal argument description  a few times

modules/file/doc.rst
```diff
@@ -213,3 +261,11 @@ path set to `/var`, the search will be activated for the current directory.
 Unless the value of `maxdepth` indicates that the search should not go beyond a
 certain number of subdirectories, and that number is reached. In which case,
 the search is deactivated.
+
+Note on symbolic links
+~~~~~~~~~~~~~~~~~~~~~~
+
+FM does not follow directory links but will follow file links. Directory links
+could lead FM to scan a path that is far out of its initial search scope, and
+can also lead to loops. A warning will be stored in the results when a directory
```
gdestuynder: warning++

modules/file/file.go
```diff
@@ -132,8 +142,11 @@ func (s *search) makeChecks() (err error) {
...
 	}()
+	if s.Options.Debug == "print" {
+		debug = true
+	}
 	if s.Options.MaxDepth == 0 {
-		s.Options.MaxDepth = 10000
+		s.Options.MaxDepth = 1000
 	}
 	if s.Options.MatchLimit == 0 {
 		s.Options.MatchLimit = 1000
```
gdestuynder: just realized we had that here, even thus its not part of the patch: im wondering if that could lead to confusion as 1000 is a great number of sub dirs, but a potentially low number of matches /offtopic

modules/file/file.go
```diff
@@ -1032,6 +1056,60 @@ func (r *run) evaluateFile(file string) (err error) {
 	return
 }
 
+/* wantThis() implements boolean logic to decide if a given check should be a match or not
```
gdestuynder: wondering if this could be done with bitwise operators for clarity (wantThis). Not sure what is clearer to most people though.

gdestuynder:
r+ for me

modules/file/file.go
jvehent: Golang doesn't have bitwise operators on booleans :/

modules/file/file.go
jvehent: That's 1,000 files returns in the results of a single search on a single host. It's possibly that people will run into this limit, but my guess is that it won't be intentional and only be the result of very large fingers :)

PR #87, opened 2015-09-19
Fix various DB Query issues

jvehent:
@ameihm0912 : r?


ameihm0912:
It looks good to me, except it looks like the defer call for rows.Close() should happen after we know we have a valid rows or statement value.

jvehent:
I fixed the `defer rows.Close()` to only happen if `rows != nil`.
I also added a fairly beefy patch to build smarter search queries, in the hope that it will improve search perfs.

PR #89, opened 2015-09-28
[minor/bug] under scheduler load dequeue command batches periodically

ameihm0912:
This resolves an issue where large volumes of commands being processed
at the same time could result in the commands being dropped by the
scheduler, due to overflows in the parameter lists during SQL statement
creation.

@jvehent r?


jvehent:
Travis seems happy. And the code looks good. I'd say r+ and will move it to prod tomorrow morning.

PR #90, opened 2015-09-28
[minor] honor os proxy environment variables in client

ameihm0912:
When building client transport, include OS related proxy configuration so client can make use of it


jvehent:
I dunno... that looks like a complex change :stuck_out_tongue_closed_eyes: 


r+

PR #130, opened 2015-10-07
[minor] add optional hashing for file module results

ameihm0912:
If the -hash option is present, include the SHA256 sum of the matched
file as part of the file info.

This is intended to be used for example with broad searches as a source for integrity monitoring across multiple runs.

### DIFF
--- a/modules/file/file.go
+++ b/modules/file/file.go
@@ -95,14 +95,14 @@ type search struct {
 }
 
 type options struct {
-	MaxDepth   float64  `json:"maxdepth"`
-	RemoteFS   bool     `json:"remotefs,omitempty"`
-	MatchAll   bool     `json:"matchall"`
-	Macroal    bool     `json:"macroal"`
-	Mismatch   []string `json:"mismatch"`
-	MatchLimit float64  `json:"matchlimit"`
-	Debug      string   `json:"debug,omitempty"`
-	Hash       bool     `json:"hash"`
+	MaxDepth     float64  `json:"maxdepth"`
+	RemoteFS     bool     `json:"remotefs,omitempty"`
+	MatchAll     bool     `json:"matchall"`
+	Macroal      bool     `json:"macroal"`
+	Mismatch     []string `json:"mismatch"`
+	MatchLimit   float64  `json:"matchlimit"`
+	Debug        string   `json:"debug,omitempty"`
+	ReturnSHA256 bool     `json:"returnsha256,omitempty"`
 }
 
 type checkType uint64
@@ -1533,7 +1533,7 @@ type fileinfo struct {
 	Size   float64 `json:"size"`
 	Mode   string  `json:"mode"`
 	Mtime  string  `json:"lastmodified"`
-	SHA256 string  `json:"sha256"`
+	SHA256 string  `json:"sha256,omitempty"`
 }
 
 // newResults allocates a Results structure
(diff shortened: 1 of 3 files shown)


modules/file/file.go
```diff
@@ -1529,9 +1530,10 @@ type matchedfile struct {
 }
 
 type fileinfo struct {
-	Size  float64 `json:"size"`
-	Mode  string  `json:"mode"`
-	Mtime string  `json:"lastmodified"`
+	Size   float64 `json:"size"`
+	Mode   string  `json:"mode"`
+	Mtime  string  `json:"lastmodified"`
+	SHA256 string  `json:"sha256"`
```
jvehent: + `omitempty`

modules/file/file.go
```diff
@@ -102,6 +102,7 @@ type options struct {
 	Mismatch   []string `json:"mismatch"`
 	MatchLimit float64  `json:"matchlimit"`
 	Debug      string   `json:"debug,omitempty"`
+	Hash       bool     `json:"hash"`
```
jvehent: `gethash` or `returnhash` would be more explicit

modules/file/file.go
```diff
@@ -1735,8 +1743,12 @@ func (r *run) PrintResults(result modules.Result, foundOnly bool) (prints []stri
 				}
 				out = fmt.Sprintf("0 match found in search '%s'", label)
 			} else {
-				out = fmt.Sprintf("%s [lastmodified:%s, mode:%s, size:%.0f] in search '%s'",
-					mf.File, mf.FileInfo.Mtime, mf.FileInfo.Mode, mf.FileInfo.Size, label)
+				out = fmt.Sprintf("%s [lastmodified:%s, mode:%s, size:%.0f",
+					mf.File, mf.FileInfo.Mtime, mf.FileInfo.Mode, mf.FileInfo.Size)
+				if mf.FileInfo.SHA256 != "" {
+					out += fmt.Sprintf(", hash:%s", strings.ToLower(mf.FileInfo.SHA256))
```
jvehent: rename `hash` with `sha256`, more explicit

modules/file/paramscreator.go
```diff
@@ -78,12 +78,14 @@ Options
 %smatchlimit <int>	- limit the number of files that can be matched by a search.
 			  the default limit is set to 1000. search will stop once the limit
 			  is reached.
+%shash			- include sha256 hash for matched files.
```
jvehent: `getsha256` would be more explicit

modules/file/paramscreator.go
```diff
@@ -394,9 +402,9 @@ func (r *run) ParamsParser(args []string) (interface{}, error) {
 		err error
 		paths, names, sizes, modes, mtimes, contents, md5s, sha1s, sha256s,
 		sha384s, sha512s, sha3_224s, sha3_256s, sha3_384s, sha3_512s, mismatch flagParam
-		maxdepth, matchlimit                 float64
-		matchall, matchany, macroal, verbose bool
-		fs                                   flag.FlagSet
+		maxdepth, matchlimit                       float64
+		hash, matchall, matchany, macroal, verbose bool
```
jvehent: same renaming

jvehent:
Ok, so if I understand this correctly, it's meant to calculate the sha256 sum of a file that matches the search criteria, and return it with the search results?
r+ on the concept and the logic. I'd like to rename the options to something more explicit though.

ameihm0912:
@jvehent r?

modules/file/paramscreator.go
```diff
@@ -78,12 +78,14 @@ Options
 %smatchlimit <int>	- limit the number of files that can be matched by a search.
 			  the default limit is set to 1000. search will stop once the limit
 			  is reached.
+%sreturnsha256		- include sha256 hash for matched files.
+			  ex: -returnsha256
```
jvehent: the dash in `ex: -returnsha256` should be parameterized

jvehent:
Looks good, but could you add paramscreator support as well. I'd like to keep parity between cmdline and console.

jvehent:
I try reading again with my :eyeglasses: on and found all the parsers...
r+

PR #133, opened 2015-10-09
[doc] Added example to look for a user's processes/connected

gdestuynder:
NOTE @jvehent : This needs a new generation of the html file as well, since as usual - we don't have exactly the same versions/files for the rendered and I don't wanna clobber everything.

(if you have a solution for that.. ;)

### DIFF
--- a/doc/cheatsheet.rst
+++ b/doc/cheatsheet.rst
@@ -117,6 +117,18 @@ return files that have unknown keys.
 	-content "^((#.+)|(\s+)?|(ssh-rsa AAAAB3NznoMzq\+2r2Vx2bhFWMU3Uuid 1061157)|(ssh-rsa AAYWH\+0XAASw== ffxbld_rsa))$" \
 	-macroal -mismatch content
 
+Find is a user is currently running any process (or is connected)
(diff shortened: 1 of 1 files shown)


doc/cheatsheet.rst
```diff
@@ -117,6 +117,18 @@ return files that have unknown keys.
...
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+If you know the UID of a user, you can check if he has any process running.
+Additionally, this means that you can find out if he's connected as well, with
+the same command.  In this example `1663` is the UID of the user we're looking
+for.
+
+.. code:: bash
+
+        mig file -path /proc/ -maxdepth 2 -name "^status$" -content "^Uid:.*1663"
```
jvehent: Just a nit, but the regex can be better written as `^Uid:\s+(1664)\s+`

jvehent:
I'm at the mercy of rst2html5, and it's non-deterministic way to generate css classes, so for now, generation of html is on me :)

doc/cheatsheet.rst
mvanotti: Typo: Find **if** a user

gdestuynder:
Thanks, updated! (note 1663 is my actual uid. i wonder who's the lucky 1664 ;)

jvehent:
Approved!

Issue #132, opened 2015-10-08
mig unable to find gpg key to create configuration file


gene1wood:
I'm not sure the method mig uses to find the gpg key but it might be extended to read the gpg conf file and use the key location indicated in the conf.

It also might be good to improve the error message to clarify that "no suitable key found" means something like

"Unable to find GPG key after searching in these places. To manually set your gpg key do this"

```
mig file --path /home -path /root -name "^authorized_keys" -content "foo"
no configuration file found at /home/jdoe/.migrc
would you like to generate a new configuration file at /home/jdoe/.migrc? Y/n> y
ReadConfiguration() -> MakeConfiguration() -> no suitable key found
```

jvehent:
Have you tried editing the location of the gpg home in ~/.migrc to resolve the issue?
Agreed on the error message: it should be improved.

ameihm0912:
Looks like a fix for this issue was merged a while ago, closing this

PR #134, opened 2015-10-09
[client] better description on gpg key not found

mvanotti:
When the .migrc is missing, a new one can be created, and if gpg is
installed, but there is no suitable key in the secring file, it will
display an error saying "No suitable key found". This commit also adds
to the error description the file which was inspected.

This fixes issue #132.

### DIFF
--- a/client/client.go
+++ b/client/client.go
@@ -217,7 +217,7 @@ func MakeConfiguration(file string) (err error) {
 		}
 	}
 	if cfg.GPG.KeyID == "" {
-		panic("no suitable key found")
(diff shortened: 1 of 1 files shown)


jvehent:
r+, but could you update your commit message to `[minor] better description on gpg key not found in client` ?
https://github.com/mozilla/mig/blob/master/CONTRIBUTING.md

mvanotti:
Should we merge this PR?

jvehent:
Eh, sorry, it slipped through the cracks.

PR #136, opened 2015-10-17
[minor] fix permissions bug on receiveModuleResults in mig-agent

mvanotti:
There was an ioutil.WriteFile whose permissions were not in octal, giving
the resulting file bad permissions.

This commit also changes some of the printf formats from '%s' to %q,
which already quotes the desired variable.


jvehent:
r+

PR #137, opened 2015-10-17
[minor] panic on agent init err if foreground mode

mvanotti:
The comment said "if in foreground mode, don't retry, just panic", but
the if condition only checked for debug, no foreground.
debug -> foreground, but you can run the mig agent in foreground without
debug mode.


jvehent:
Good catch. r+
fyi, for later commits, please add `/bug` when it fixes a bug. such as `[minor/bug] some commit message`

PR #138, opened 2015-10-17
[minor] fix configLoad bug in mig-agent

mvanotti:
If an error ocurred during configLoad function, it might not use the
builtin conf, but a half-overwritten one. This change defers the
Config overwrite until all the error-sensitive functions are called.

### DIFF
--- a/mig-agent/agent.go
+++ b/mig-agent/agent.go
@@ -123,9 +123,10 @@ func main() {
 	// attempt to read a local configuration file
 	err = configLoad(*config)
 	if err != nil {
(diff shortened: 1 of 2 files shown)


jvehent:
it is cleaner indeed. I guess we never really ran into that bug.

jvehent:
I had to revert the `[error]` into ` [warn]` in 3cfb4a42207ec4b9135de55c25388b9d85d86820 . Displaying an error makes people think that something's wrong, when it's just the normal behavior of the agent.

PR #140, opened 2015-10-19
Use servers in /etc/ntp.conf for timedrift

### LINKED ISSUE #116: [mig modules] timedrift should read target ntp server from /etc/ntp.conf (Bugzilla #1149496)
Migrated from https://bugzilla.mozilla.org/show_bug.cgi?id=1149496

On 2015-03-31 05:45:42 -0700, Julien Vehent [:ulfr] wrote:

> In some instances, endpoints cannot establish NTP connections with pools from the internet. Timedrift should attempt to connect to locally configured NTP servers first, by reading /etc/ntp.conf and parsing the ^server entries, and then use internet servers if no local ones are available.

eXeC64:
This pull request fixes #116. Instead of randomly picking an ntp server out of a list of hardcoded ones we now try all the servers in `/etc/ntp.conf` first, and then move onto the hardcoded list.

### DIFF
--- a/modules/timedrift/timedrift.go
+++ b/modules/timedrift/timedrift.go
@@ -20,6 +20,7 @@ import (
 	"mig.ninja/mig/modules"
 	"net"
 	"os"
+	"strings"
 	"time"
 )
 
@@ -64,6 +65,13 @@ type ntpstats struct {
 	Reachable bool      `json:"reachable"`
 }
 
+var NtpBackupPool = []string{
+	`time.nist.gov`,
+	`0.pool.ntp.org`,
+	`1.pool.ntp.org`,
+	`2.pool.ntp.org`,
+	`3.pool.ntp.org`}
+
 func (r *run) ValidateParameters() (err error) {
 	if r.Parameters.Drift != "" {
 		_, err = time.ParseDuration(r.Parameters.Drift)
@@ -73,9 +81,12 @@ func (r *run) ValidateParameters() (err error) {
 
 func (r *run) Run(in io.Reader) (out string) {
 	var (
-		stats statistics
-		el    elements
-		drift time.Duration
+		stats   statistics
+		el      elements
+		drift   time.Duration
+		ntpFile *os.File
+		ntpScan *bufio.Scanner
+		ntpPool []string
 	)
 	defer func() {
 		if e := recover(); e != nil {
@@ -106,12 +117,32 @@ func (r *run) Run(in io.Reader) (out string) {
 	}
 	// assume host has synched time and set to false if not true
 	el.IsWithinDrift = true
+
+	//Load ntp servers from /etc/ntp.conf
+	ntpFile, err = os.Open("/etc/ntp.conf")
+	if err != nil {
+		r.Results.Errors = append(r.Results.Errors,
+			fmt.Sprintf("Using backup NTP hosts. Failed to read /etc/ntp.conf with error '%v'", err))
+	} else {
+		defer ntpFile.Close()
+		ntpScan = bufio.NewScanner(ntpFile)
+		for ntpScan.Scan() {
+			ntpFields := strings.Fields(ntpScan.Text())
+			if len(ntpFields) < 2 {
+				continue
+			}
+			if ntpFields[0] == "server" {
+				ntpPool = append(ntpPool, ntpFields[1])
+			}
+		}
+	}
+
+	//Add our hardcoded online servers to the end of our ntpPool as fallbacks
+	ntpPool = append(ntpPool, NtpBackupPool...)
+
 	// attempt to get network time from each of the NTP servers, and exit
 	// as soon as we get a valid result from one of them
-	for i := 0; i < len(NtpPool); i++ {
-
-		// pick a server between 0 and len of ntppool, somewhat randomly
-		ntpsrv := NtpPool[time.Now().Nanosecond()%len(NtpPool)]
+	for _, ntpsrv := range ntpPool {
 		t, lat, err := GetNetworkTime(ntpsrv)
 		if err != nil {
 			// failed to get network time, log a failure and try another one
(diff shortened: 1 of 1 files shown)


modules/timedrift/timedrift.go
```diff
@@ -106,12 +118,31 @@ func (r *run) Run(in io.Reader) (out string) {
 	}
 	// assume host has synched time and set to false if not true
 	el.IsWithinDrift = true
+
+	//Load ntp servers from /etc/ntp.conf
+	ntpFile, err = os.Open("/etc/ntp.conf")
+	if err != nil {
+		panic(err)
```
jvehent: The module should just continue with the backup ntp hosts if `/etc/ntp.conf` is unreadable, and store it into the returned errors:
```go
r.Results.Errors = append(r.Results.Errors, fmt.Sprintf("Using backup NTP hosts. Failed to read /etc/ntp.conf with error '%v'", err))
```

modules/timedrift/timedrift.go
```diff
@@ -106,12 +118,31 @@ func (r *run) Run(in io.Reader) (out string) {
...
+
+	ntpScan = bufio.NewScanner(ntpFile)
+	for ntpScan.Scan() {
+		ntpLine := strings.TrimSpace(ntpScan.Text())
+		if strings.HasPrefix(ntpLine, "server ") {
+			NtpPool = append(NtpPool, strings.TrimSpace(ntpLine[7:]))
+		}
+	}
+	if err = ntpScan.Err(); err != nil {
+		panic(err)
```
jvehent: Same as above: store the error and continue with the backup hosts.

modules/timedrift/timedrift.go
```diff
@@ -106,12 +118,31 @@ func (r *run) Run(in io.Reader) (out string) {
...
+	if err != nil {
+		panic(err)
+	}
+	defer ntpFile.Close()
+
+	ntpScan = bufio.NewScanner(ntpFile)
+	for ntpScan.Scan() {
+		ntpLine := strings.TrimSpace(ntpScan.Text())
+		if strings.HasPrefix(ntpLine, "server ") {
+			NtpPool = append(NtpPool, strings.TrimSpace(ntpLine[7:]))
```
jvehent: You should check that ntpLine has a length >  7.

jvehent: Looks like the string needs to be split a little further: the format is `server address [options ...]` and we just want the address.
Debian ships with these lines:
```
server 0.debian.pool.ntp.org iburst
server 1.debian.pool.ntp.org iburst
server 2.debian.pool.ntp.org iburst
server 3.debian.pool.ntp.org iburst
```
A `string.Split` that extracts the 2nd entry should do the job.

jvehent:
I like it! Thanks a lot for the patch! It has a few minor issues, but overall the code is good.

jvehent:
Test with:
```bash
$ ./bin/linux/amd64/mig-agent-latest -p -m timedrift <<< '{"class":"parameters","parameters":{"drift":"60s"}}'
[error] Failed to load conf from "/etc/mig/mig-agent.cfg", err: configLoad() -> open /etc/mig/mig-agent.cfg: no such file or directory
[info] Using builtin conf.
local time is 2015-10-19T08:00:57.065480118-04:00
local time is within acceptable drift from NTP servers
stat: execution time was 298.300033ms
stat: 0.debian.pool.ntp.org iburst was unreachable
stat: 1.debian.pool.ntp.org iburst was unreachable
stat: 2.debian.pool.ntp.org iburst was unreachable
stat: 3.debian.pool.ntp.org iburst was unreachable
stat: time.nist.gov responded in 215.626792ms with time 2015-10-19 12:00:57.203915249 +0000 UTC. local time drifts by 159.924127ms
timedrift module has succeeded
```

eXeC64:
Thanks for the testing guidance, it was very helpful in revising the patch.

Changes:
 * Moved NtpPool to ntpPool, a local variable. It had no reason to be global any more.
 * Reworked the `/etc/ntp.conf` parsing to use strings.Fields (Which also adds support for tab separation)
 * Report the failure to open `/etc/ntp.conf` with `r.Results.Errors`

There is one thing I'm unsure about currently, which is the behaviour when `/etc/ntp.conf` cannot be read. Currently it will report a failure, even though using a backup server succeeded. This seems to be because an error was reported. Is there a way to report it as a warning not an error? That seems more suitable to this type of error.

jvehent:
It looks good!
```bash
$ ./bin/linux/amd64/mig-agent-latest -m timedrift <<< '{"class":"parameters","parameters":{"drift":"60s"}}'|j
[warn] Could not load a local conf from "/etc/mig/mig-agent.cfg", err: configLoad() -> open /etc/mig/mig-agent.cfg: no such file or directory
[info] Using builtin conf.
{
    "elements": {
        "hascheckeddrift": true,
        "iswithindrift": true,
        "localtime": "2015-10-19T12:22:03.871107071-04:00"
    },
    "errors": null,
    "foundanything": false,
    "statistics": {
        "exectime": "45.67901ms",
        "ntpstats": [
            {
                "drift": "22.649704ms",
                "host": "0.debian.pool.ntp.org",
                "latency": "43.344918ms",
                "reachable": true,
                "time": "2015-10-19T12:22:03.894204172-04:00"
            }
        ]
    },
    "success": true
}
```

On the error reporting: I agree the current naming is not optimal... `modules.Errors []string` can contain a mix of failures and warning, and we use `modules.Success bool` check that it succeeded or failed. It's not great, but that's all we have for now, so your code is fine.
source: http://mig.mozilla.org/doc/modules.rst.html#success

modules/timedrift/timedrift.go
```diff
@@ -64,6 +65,13 @@ type ntpstats struct {
 	Reachable bool      `json:"reachable"`
 }
 
+var NtpBackupPool = []string{
+	`time.nist.gov`,
```
ameihm0912: Is there any reason we are treating these as raw string literals vs using a standard quote?

Not a big deal, just wondering.

ameihm0912:
r+

jvehent:
Well then, it's a merge! Thanks a lot @eXeC64 , and welcome to the project. I hope you have more of those coming ;)

PR #141, opened 2015-10-19
[minor] Mark and ignore bad files in mig-scheduler. Fixes #131.

### LINKED ISSUE #131: [mig scheduler] Invalid data in scheduler returned directory
If valid JSON is returned to the scheduler that is missing some fields, the scheduler will not load it (CmdFromFile() fails). The data ends up sitting in the returned directory though and is reparsed every few seconds. If a number of invalid events were returned this could potentially get inefficient and may cause problems.

We could somehow mark a failed returned file, so the first time it fails maybe it's renamed, and subsequent walks of the spool directory ignore these failed files.

jdiez17:


jvehent:
Thanks for the patch. The build doesn't pass because of this error:
```
mig-scheduler/collector.go:138: undefined: strings in strings.Index
```
You forgot to import "strings" :)

mig-scheduler/collector.go
```diff
@@ -134,6 +134,12 @@ func loadReturnedCommands(ctx Context) (err error) {
 			continue
 		}
 		filename := ctx.Directories.Command.Returned + "/" + DirEntry.Name()
+
+		if strings.Index(filename, ".fail") == len(filename)-5 {
```
jvehent: Why not just use `strings.HasSuffix()` ? 
https://golang.org/pkg/strings/#HasSuffix

jvehent:
Thanks for the patch. It is indeed most welcomed! If you could just fix the 2 comments I left, I'll merge it.

jdiez17:
Thanks, fixed those two issues. I force-pushed my branch to rebase the two commits as per CONTRIBUTING.md. Should be ready to merge.

ameihm0912:
r+, Thanks for grabbing this!

jvehent:
Absolutely! Welcome to the project @jdiez17 ! Keep the patches flowing :)

PR #142, opened 2015-10-21
[minor] detect/interrogate linux netns namespaces in netstat module

ameihm0912:
This patch adds support for network namespace switching to the netstat
module, intended for interrogating environments such as docker
environments from a privileged host running an agent.

### DIFF
--- a/modules/netstat/netstat.go
+++ b/modules/netstat/netstat.go
@@ -44,6 +44,7 @@ type params struct {
 	NeighborIP    []string `json:"neighborip,omitempty"`
 	ConnectedIP   []string `json:"connectedip,omitempty"`
 	ListeningPort []string `json:"listeningport,omitempty"`
+	Namespaces    bool     `json:"namespaces,omitempty"`
 }
 
 type elements struct {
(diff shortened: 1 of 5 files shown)


ameihm0912:
@gdestuynder r? on approach ?

gdestuynder:
If I understand this correctly, this is made so that a namespace can interrogate another - since the host namespace will always be able to see and query all other namespaces anyway (if you're in the host namespace, you don't need to setns(), you just crawl through /proc/<pid> and associate ns id's to tcp/udp/etc tables, like netstat would, and search in that).

However, this would also mean that this only works in an environment where the namespace you're in is able to transition back to another and/or the host namespace. This only works if you are in the same PID namespace (or have a reference to the host's ns file reference but I suspect it's impossible since it's only in /proc, which is managed by the kernel and displays whatever PID namespace you're in).

i.e. I'm wondering if the setns is necessary at all, albeit I may have missed something

ameihm0912:
After discussion with @gdestuynder we are going to close this and approach a different way as suggested. Based on some of the incoming sandboxing work we want to avoid conditional system calls in certain situations as this will increase the complexity associated with sandboxing the module.

PR #145, opened 2015-10-30
namespace support in netstat module

ameihm0912:
@gdestuynder @jvehent please review when possible

This patch adds support to the netstat module to identify network activity in other namespaces granted the agent is running with access to a global PID namespace. Currently in the netstat module, MIG can only see network activity related to the namespace it runs in.

This support has been added as an option, to maintain the existing behavior. Ideally we would change it to the default at some point.

One caveat is local interface MAC detection does not support this new functionality as we are using go net.Interfaces() here vs. inspecting the system directly in HasLocalMAC(). This function would need to be changed to support this. However other MAC detection (e.g., neighbor MAC) works as expected.

### DIFF
--- a/modules/netstat/netstat.go
+++ b/modules/netstat/netstat.go
@@ -38,12 +38,13 @@ type run struct {
 }
 
 type params struct {
-	LocalMAC      []string `json:"localmac,omitempty"`
-	LocalIP       []string `json:"localip,omitempty"`
-	NeighborMAC   []string `json:"neighbormac,omitempty"`
-	NeighborIP    []string `json:"neighborip,omitempty"`
-	ConnectedIP   []string `json:"connectedip,omitempty"`
-	ListeningPort []string `json:"listeningport,omitempty"`
+	LocalMAC         []string `json:"localmac,omitempty"`
+	LocalIP          []string `json:"localip,omitempty"`
+	NeighborMAC      []string `json:"neighbormac,omitempty"`
+	NeighborIP       []string `json:"neighborip,omitempty"`
+	ConnectedIP      []string `json:"connectedip,omitempty"`
+	ListeningPort    []string `json:"listeningport,omitempty"`
+	SearchNamespaces bool     `json:"namespaces,empty"`
 }
 
 type elements struct {
@@ -62,6 +63,7 @@ type element struct {
 	LocalPort     float64 `json:"localport,omitempty"`
 	RemoteAddr    string  `json:"remoteaddr,omitempty"`
 	RemotePort    float64 `json:"remoteport,omitempty"`
+	Namespace     string  `json:"namespace,omitempty"`
 }
 
 func newElements() *elements {
@@ -83,6 +85,11 @@ type statistics struct {
 	Exectime  string  `json:"exectime"`
 	Totalhits float64 `json:"totalhits"`
 }
+
+// Indicates module should use namespace resolution mode, which changes
+// the behavior on Linux; optional right now but may be the default in
+// the future.
+var namespaceMode bool
 
 func (r *run) ValidateParameters() (err error) {
 	for _, val := range r.Parameters.LocalMAC {
@@ -175,6 +182,10 @@ func (r *run) Run(in io.Reader) (resStr string) {
 	err = r.ValidateParameters()
 	if err != nil {
 		panic(err)
+	}
+
+	if r.Parameters.SearchNamespaces {
+		namespaceMode = true
 	}
 
 	els := *newElements()
@@ -319,6 +330,13 @@ func HasLocalIP(ipStr string) (found bool, elements []element, err error) {
 	return
 }
 
+func printNamespaceId(ns string) string {
+	if ns != "" && ns != "default" {
+		return fmt.Sprintf(" namespace:[%v]", ns)
+	}
+	return ""
+}
+
 func (r *run) PrintResults(result modules.Result, matchOnly bool) (prints []string, err error) {
 	var (
 		el    elements
@@ -340,6 +358,7 @@ func (r *run) PrintResults(result modules.Result, matchOnly bool) (prints []stri
 		}
 		for _, el := range res {
 			resStr := fmt.Sprintf("found local mac %s for netstat localmac:'%s'", el.LocalMACAddr, val)
+			resStr += printNamespaceId(el.Namespace)
 			prints = append(prints, resStr)
 		}
 	}
@@ -350,6 +369,7 @@ func (r *run) PrintResults(result modules.Result, matchOnly bool) (prints []stri
 		for _, el := range res {
 			resStr := fmt.Sprintf("found neighbor mac %s %s for netstat neighbormac:'%s'",
 				el.RemoteMACAddr, el.RemoteAddr, val)
+			resStr += printNamespaceId(el.Namespace)
 			prints = append(prints, resStr)
 		}
 		if len(res) == 0 {
@@ -363,6 +383,7 @@ func (r *run) PrintResults(result modules.Result, matchOnly bool) (prints []stri
 		}
 		for _, el := range res {
 			resStr := fmt.Sprintf("found local ip %s for netstat localip:'%s'", el.LocalAddr, val)
+			resStr += printNamespaceId(el.Namespace)
 			prints = append(prints, resStr)
 		}
 		if len(res) == 0 {
@@ -377,6 +398,7 @@ func (r *run) PrintResults(result modules.Result, matchOnly bool) (prints []stri
 		for _, el := range res {
 			resStr := fmt.Sprintf("found connected tuple %s:%.0f with local tuple %s:%.0f for netstat connectedip:'%s'",
 				el.RemoteAddr, el.RemotePort, el.LocalAddr, el.LocalPort, val)
+			resStr += printNamespaceId(el.Namespace)
 			prints = append(prints, resStr)
 		}
 		if len(res) == 0 {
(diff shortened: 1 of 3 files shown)


gdestuynder:
First looks looks good!
Gotta have second look this weekend or monday :)

modules/netstat/netstat.go
```diff
@@ -84,6 +86,11 @@ type statistics struct {
 	Totalhits float64 `json:"totalhits"`
 }
 
+// Indicates module should use namespace resolution mode, which changes
+// the behavior on Linux; optional right now but may be the default in
+// the future.
+var namespaceMode bool
```
gdestuynder: it makes sense to me for that to be default since we ignore if namespace cant be read.
this is only optional because its new i guess right?

gdestuynder:
added comment otherwise r+

modules/netstat/netstat.go
ameihm0912: That's correct, I just didn't want to initially replace the existing method. My thoughts were we support both and remove the previous method later.

ameihm0912:
@jvehent you OK with a merge here?

jvehent:
I'll review in the morning.

modules/netstat/netstat.go
gdestuynder: ok sounds good in this case

jvehent:
nit: rebase all 3 commits into a single `medium` level commit

modules/netstat/netstat.go
```diff
@@ -44,6 +44,7 @@ type params struct {
 	NeighborIP    []string `json:"neighborip,omitempty"`
 	ConnectedIP   []string `json:"connectedip,omitempty"`
 	ListeningPort []string `json:"listeningport,omitempty"`
+	Namespaces    bool     `json:"namespaces,empty"`
```
jvehent: nit: maybe `SearchNamespaces` would be more explicit?

modules/netstat/netstat.go
```diff
@@ -62,6 +63,7 @@ type element struct {
 	LocalPort     float64 `json:"localport,omitempty"`
 	RemoteAddr    string  `json:"remoteaddr,omitempty"`
 	RemotePort    float64 `json:"remoteport,omitempty"`
+	Namespace     string  `json:"namespace,omitempty"`
```
jvehent: It seems to be a uint value. any reason to store it as a string?
`NamespaceId` would be more explicit.

modules/netstat/netstat.go
```diff
@@ -319,6 +330,13 @@ func HasLocalIP(ipStr string) (found bool, elements []element, err error) {
 	return
 }
 
+func resultNamespace(ns string) string {
```
jvehent: nit: `func printNamespaceId()` ?

modules/netstat/netstat.go
```diff
@@ -340,6 +358,7 @@ func (r *run) PrintResults(result modules.Result, matchOnly bool) (prints []stri
 		}
 		for _, el := range res {
 			resStr := fmt.Sprintf("found local mac %s for netstat localmac:'%s'", el.LocalMACAddr, val)
+			resStr += resultNamespace(el.Namespace)
```
jvehent: could also be written as one line
```go
resStr := fmt.Sprintf("found local mac %s for netstat localmac:'%s'%s", el.LocalMACAddr, val, resultNamespace(el.Namespace))
```

jvehent: meh. it doesn't look much prettier. scratch that comment.

modules/netstat/netstat_linux.go
```diff
@@ -9,15 +9,152 @@ package netstat /* import "mig.ninja/mig/modules/netstat" */
...
 )
 
-// HasSeenMac on linux looks for a matching mac address in /proc/net/arp
-// and returns its MAC and IP address if found
+// Represents lines obtained from the /proc file system related to network
+// activity
+type procNetLine struct {
+	line         string // The line from the proc file
+	nsIdentifier string // A namespace identifier associated with the process
+}
```
jvehent: I don't really understand what this does. Could you add a line example in the comment?

modules/netstat/netstat_linux.go
```diff
@@ -9,15 +9,152 @@ package netstat /* import "mig.ninja/mig/modules/netstat" */
...
+
+// Return process candidates organized by active network namespace, this will
+// include the host namespace in addition to other namespaces
+func procCandidates() (map[string][]string, error) {
+	var err error
+	ret := make(map[string][]string)
```
jvehent: so, if I understand correctly, this is a map of ns ids, where each ns has a list of processes. correct?
```json
{
  "4026531957": [ "/proc/1/net", "/proc/123/net" ],
   "8716321": [ "/proc/42/net", "/proc/51/net" ]
}
```

modules/netstat/netstat_linux.go
```diff
@@ -9,15 +9,152 @@ package netstat /* import "mig.ninja/mig/modules/netstat" */
...
+	}
+
+	return ret, nil
+}
+
+// Read lines from a network proc file; used by procNetNS()
+func procNetNSEntries(fname string) (ret []string, err error) {
+	fd, err := os.Open(fname)
+	if err != nil {
+		return ret, err
```
jvehent: nit: no need to `return ret, err`, just `return` since the returned vars are defined in the function signature.

jvehent:
I know it's a C thing, but I find it hard to review code that's organize bottom-up. All of mig's code is organized top-down, with the lower functions placed right after the upper function that calls them. It makes reading code more natural, and I'd very much like to follow this logic in this patch too.

modules/netstat/netstat_linux.go
```diff
@@ -9,15 +9,152 @@ package netstat /* import "mig.ninja/mig/modules/netstat" */
...
+	procCache, err := procCandidates()
+	if err != nil {
+		panic(err)
+	}
+
+	// For each cached namespace, attempt to extract data from one
+	// process. We try until we find one to help work around the
+	// potential for the process to exit after we have cached the
+	// list.
+	for k, v := range procCache {
```
jvehent: nit: `k` == `nsid`; `v`==`procpaths`

modules/netstat/netstat_linux.go
```diff
@@ -9,15 +9,152 @@ package netstat /* import "mig.ninja/mig/modules/netstat" */
...
+// <pid>/net/tcp visible by the agent, return a list of proc network entries
+// for each active network namespace, including the host.
+//
+// This function works by building a cache of all current network namespaces
+// on the host based on /proc. From this list, we try to collect data from the
+// indicated file for at most one process per namespace. Since <pid>/net/*
+// represents network activity for the namespace the process is in, and not
+// the process itself we only need to be successful with one PID in each
+// namespace.
+func procNetNS(fname string) (ret []procNetLine, err error) {
```
jvehent: Two issues can happen here:

1. `procNetNSEntries` could fail toward the end of the file, and the whole file will be throw away because the last line returned an error. it would be nice to inspect what we can, even if it's incomplete data.

2. from an optimization point of view, on a busy system, this will return a huge slice of entries. We run mig on load balancers that have hundreds of thousands of connections (a single proc can easily hold ~200k sockets). A typical entry is ~120 bytes long, so that's a 22MB slice.

Go might be fine with it, and the target system too, but it's still a lot of data to pass around slices, and a bunch of unnecessary pressure on the garbage collector. I think this could be refactor to not use slices, but iterate on the scanner instead.

jvehent: What about using a channel that `procNetNS` writes `procNetLine`s into, and the caller functions can read from the scan the files? That would circumvent the need for a slice of lines.

modules/netstat/netstat_linux.go
```diff
@@ -9,15 +9,152 @@ package netstat /* import "mig.ninja/mig/modules/netstat" */
...
+	defer func() {
+		if e := recover(); e != nil {
+			err = fmt.Errorf("procArpEntries(): %v", e)
+		}
+	}()
+
+	// If we have namespace resolution enabled, use that
+	if namespaceMode {
+		return procArpEntriesNS()
+	}
```
jvehent: should namespace mode bypass reading `/proc/net/arp` entirely? Would reading both return duplicates?

jvehent:
Review done. I think this could work just fine, but if there's a way to avoid carrying large slices around, it would make the module much lighter to run.
The namespace support looks good to me.

modules/netstat/netstat.go
ameihm0912: Resolved

modules/netstat/netstat.go
ameihm0912: It's intended to be a more abstract indicator. The identifier is uint on Linux, but the idea was this field could be utilized with other platforms if we run into it in the future. Resolved name of field.

modules/netstat/netstat.go
ameihm0912: Resolved.

modules/netstat/netstat_linux.go
ameihm0912: Resolved.

modules/netstat/netstat_linux.go
ameihm0912: That's correct; it's a list of candidates for inspection, we only look at one.

modules/netstat/netstat_linux.go
ameihm0912: Resolved.

modules/netstat/netstat_linux.go
ameihm0912: So, regarding 1. this is intentional. The reason being the net/ entry represents the entire namespace, not just that process. If we keep what we have up until that point, when we try the next PID in that namespace we will need to filter out entries we have already seen.

Re: 2, one of the larger copies has been eliminated using channels. One of the reasons for this is I wanted to maintain as much of the older function implementations as possible for now. My suggestion here is we leave the remaining copies for now, and when we make this mode the default refactor it as suggested?

modules/netstat/netstat_linux.go
ameihm0912: Yeah we would get duplicates; worst case if you want that file specifically we could use the legacy mode right now.

PR #147, opened 2015-11-03
[minor/bug] fix module parameter input issues

ameihm0912:
This resolves a bug introduced in e2e687f. The args parameter to
runModuleDirectly() used to contain input parameters, but as of recent
changes to allow parameters via stdin does not.

Since the parameter was unused, this changes it to be an interface
pointer. If set, this will be used as module parameters, if not the
agent modules will expect input via stdin as normal.

### DIFF
--- a/mig-agent/agent.go
+++ b/mig-agent/agent.go
@@ -6,6 +6,7 @@
 package main
 
 import (
+	"bufio"
 	"bytes"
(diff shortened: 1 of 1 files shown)


ameihm0912:
@jvehent r?

jvehent:
r+

PR #149, opened 2015-11-04
[medium] Force max exec time for agent in checkin mode, fixes #148

### LINKED ISSUE #148: in checkin mode, limit max agent exec time to value of MODULETIMEOUT
Checkin mode should not be able to hang, ever. We could use the value of MODULETIMEOUT to kill the agent when the delay is expired.

jvehent:
This patch imposes the same max exec time to the mig agent in checkin mode that we use for modules. If the agent takes longer than MODULETIMEOUT to run, it will be force killed, regardless of pending operations.

This was testing on my dev instance by running a file search on the entire `/` of my local drive.
```
$ sudo ./bin/linux/amd64/mig-agent-latest -m agent-checkin -d
[warn] Could not load a local conf from "/etc/mig/mig-agent.cfg", err: configLoad() -> open /etc/mig/mig-agent.cfg: no such file or directory
[info] Using builtin conf.
.....
2015/11/04 10:54:17 5203168985088 - - [debug] executing module "file"
2015/11/04 10:54:20 - - - [info] No outstanding messages in relay.
2015/11/04 10:54:20 - - - [info] Agent is done checking in. waiting for all modules to complete.
[critical] Agent in checkin mode reached max exec time, exiting
```

@ameihm0912 : r?

### DIFF
--- a/mig-agent/agent.go
+++ b/mig-agent/agent.go
@@ -127,9 +127,22 @@ func main() {
 		}
 	case "agent-checkin":
 		*foreground = true
-		err = runAgentCheckin(*foreground, *upgrading, *debug)
-		if err != nil {
-			panic(err)
+		// in checkin mode, the agent is not allowed to run for longer than
+		// MODULETIMEOUT, so we create a timer that force exit if the run
+		// takes too long.
+		done := make(chan error, 1)
+		go func() {
+			done <- runAgentCheckin(*foreground, *upgrading, *debug)
+		}()
+		select {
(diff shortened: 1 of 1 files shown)


jvehent:
@mvanotti : we really want a hard kill when the timeout is reached, so calling destroy would induce some uncertainty if the call blocks. I haven't looked into Context, but it seems like a good approach with should standardize on for modules as well. Maybe something for a bigger refactoring?

ameihm0912:
@jvehent looks good to me.

PR #150, opened 2015-11-05
Scheduler relay heartbeat

jvehent:
If RabbitMQ suffers a network partition, the scheduler will remain in a half broken state where no message can be consumed or published, but other goroutines work fine. This patch introduce a new dummy event that heartbeats the scheduler and terminates it if the publication of the heartbeat fails. The hearbeat is a normal event that expires in rabbitmq after 100 minutes, like any other event. It is published to a queue that could eventually be consumed for monitoring.

r? @ameihm0912

### DIFF
--- a/constants.go
+++ b/constants.go
@@ -18,4 +18,7 @@ const (
 	Ev_Q_Agt_Auth_Fail = "agent.authentication.failure"
 	Ev_Q_Agt_New       = "agent.new"
 	Ev_Q_Cmd_Res       = "command.results"
+
+	// dummy queue for scheduler heartbeats to the relays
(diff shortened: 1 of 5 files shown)


ameihm0912:
Looks good to me. Would we ever run into a situation where sendEvent() succeeds but the message doesn't actually make it to the relay? For example, will the AMQP library queue the events for submission and return success. I guess we can resolve that scenario by monitoring for the message.

jvehent:
It's possible. I haven't looked into the amqp library code enough to say that it will queue at that level. From my tests, restarting rabbitmq is sufficient to drop the scheduler connection and make it fail the heartbeat, so it should solve our immediate problem as it is.

PR #153, opened 2015-11-16
[minor] vendoring update

jvehent:
Anyone sees anything suspicious before we merge it?


vendor/golang.org/x/crypto/openpgp/packet/symmetric_key_encrypted.go
```diff
@@ -66,42 +64,35 @@ func (ske *SymmetricKeyEncrypted) parse(r io.Reader) (err error) {
...
-// ske.Key will contain the session key.
-func (ske *SymmetricKeyEncrypted) Decrypt(passphrase []byte) error {
-	if !ske.Encrypted {
-		return nil
-	}
-
+// Decrypt attempts to decrypt an encrypted session key and returns the key and
+// the cipher to use when decrypting a subsequent Symmetrically Encrypted Data
+// packet.
+func (ske *SymmetricKeyEncrypted) Decrypt(passphrase []byte) ([]byte, CipherFunction, error) {
```
ameihm0912: Does this change break code in pgp/sign.go?

e.g. ./pgp/sign.go:		err = signer.PrivateKey.Decrypt([]byte(cachedPassphrase))

ameihm0912:
@jvehent I added one note but other then that I don't see anything that stands out, although it's a lot to go through.

Most of the changes in scribe were not functional but more related to exporting various data types in the library.

vendor/golang.org/x/crypto/openpgp/packet/symmetric_key_encrypted.go
jvehent: This change is for a different type, ours uses `func (pk *PrivateKey) Decrypt(passphrase []byte) error {`

jvehent:
Agreed, it's a lot to go through at once. We should try to do this twice a month to reduce the size of the change. Seems like a good task to start in the gator meeting :)

PR #158, opened 2015-12-16
[medium] Unify hash selection under sha2/sha3; fixes #155

### LINKED ISSUE #155: [mig modules] mig file hash function selection could be friendlier
Instead of requiring:

-sha256 <hash>
-sha384 <hash>
-sha512 <hash>
-sha3_224 <hash>
-sha3_256 <hash>
-sha3_384 <hash>
-sha3_512 <hash>

It should just use:
-sha2 <hash>
-sha3 <hash>

And then look at the hash size (256 bits, 512 bits, etc.) to see what hash function to run.

therealkbhat:
I tried fusing the hashes under sha2/sha3 as we discussed on IRC, but the tests aren't passing.

[Here](https://gist.github.com/therealkbhat/b2ffffa2e2ef7e22c9f3) is the error log.

So, this is a WIP, with the following points to be addressed:
* I need a bit of help on figuring out what's wrong with my code.
* I need to add more test cases to ensure that every way the "-sha2" and "-sha3" flags (with hashes of varying length) can be used are indeed tested.

Thanks in advance for your feedback!

### DIFF
--- a/modules/file/file.go
+++ b/modules/file/file.go
@@ -79,13 +79,8 @@ type search struct {
 	Mtimes       []string `json:"mtimes,omitempty"`
 	MD5          []string `json:"md5,omitempty"`
 	SHA1         []string `json:"sha1,omitempty"`
-	SHA256       []string `json:"sha256,omitempty"`
-	SHA384       []string `json:"sha384,omitempty"`
-	SHA512       []string `json:"sha512,omitempty"`
-	SHA3_224     []string `json:"sha3_224,omitempty"`
-	SHA3_256     []string `json:"sha3_256,omitempty"`
-	SHA3_384     []string `json:"sha3_384,omitempty"`
-	SHA3_512     []string `json:"sha3_512,omitempty"`
+	SHA2         []string `json:"sha2,omitempty"`
+	SHA3         []string `json:"sha3,omitempty"`
 	Options      options  `json:"options,omitempty"`
 	checks       []check
 	checkmask    checkType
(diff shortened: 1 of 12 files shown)


modules/file/file.go
```diff
@@ -512,51 +488,32 @@ func (r *run) ValidateParameters() (err error) {
...
-			if err != nil {
-				return
+			switch len(hash) {
+			case 64:
+				err = validateHash(hash, checkSHA256)
+			case 96:
+				err = validateHash(hash, checkSHA384)
+			case 128:
+				err = validateHash(hash, checkSHA512)
 			}
```
jvehent: there should be a default case here that fails validation

modules/file/file.go
```diff
@@ -512,51 +488,32 @@ func (r *run) ValidateParameters() (err error) {
...
-				return
+			switch len(hash) {
+			case 56:
+				err = validateHash(hash, checkSHA3_224)
+			case 64:
+				err = validateHash(hash, checkSHA3_256)
+			case 96:
+				err = validateHash(hash, checkSHA3_384)
+			case 128:
+				err = validateHash(hash, checkSHA3_512)
```
jvehent: same as above: need a default case that fails validation

jvehent:
You'll also have to patch `mig-api/compliance.go`, which references the old SHA methods to generate compliance items:
```
mig-api/compliance.go:135: mf.Search.SHA256 undefined (type file.search has no field or method SHA256)
mig-api/compliance.go:141: mf.Search.SHA384 undefined (type file.search has no field or method SHA384)
mig-api/compliance.go:147: mf.Search.SHA512 undefined (type file.search has no field or method SHA512)
mig-api/compliance.go:153: mf.Search.SHA3_224 undefined (type file.search has no field or method SHA3_224)
mig-api/compliance.go:159: mf.Search.SHA3_256 undefined (type file.search has no field or method SHA3_256)
mig-api/compliance.go:165: mf.Search.SHA3_384 undefined (type file.search has no field or method SHA3_384)
mig-api/compliance.go:171: mf.Search.SHA3_512 undefined (type file.search has no field or method SHA3_512)
```

modules/file/file_test.go
```diff
@@ -488,18 +472,12 @@ func TestMismatch(t *testing.T) {
...
-				SHA3_384: []string{TESTDATA[2].sha3_384},
-				SHA3_512: []string{TESTDATA[2].sha3_512},
+				SHA2:     []string{TESTDATA[2].sha2},
+				SHA3:     []string{TESTDATA[2].sha3},
 				Options: options{
 					MaxDepth: 1,
 					MatchAll: true,
-					Mismatch: []string{`md5`, `sha1`, `sha256`, `sha384`, `sha512`,
-						`sha3_224`, `sha3_256`, `sha3_384`, `sha3_512`},
+					Mismatch: []string{`md5`, `sha1`, `sha2`, `sha3`},
```
jvehent: This test that is failing: `file_test.go:490: want files that don't match the hashes of testfile2, should find testfile0 & 1`
The parameters used by the test are:
```json
{
	"class": "parameters",
	"parameters": {
		"searches": {
			"s1": {
				"paths": ["/tmp/migfiletest08-45-21.93044607"],
				"md5": ["8d3a7afb7e59693b383d52396243a5b8"],
				"sha1": ["d82bc1145d471714b056940b268032f9ab0df2ae"],
				"sha2": ["3b495fae5bae9751ea4706c29e992002ba277bce30bd83a827b01ba977eabc2f"],
				"sha3": ["fdb23afa808c265284c3199013e4ded9704eebf54ffdc1f016dacc12"],
				"options": {
					"maxdepth": 1,
					"matchall": true,
					"macroal": false,
					"mismatch": ["md5", "sha1", "sha2", "sha3"],
					"matchlimit": 0
				}
			}
		}
	}
}
```
And it returns:
```json
{
	"foundanything": true,
	"success": true,
	"elements": {
		"s1": [{
			"file": "",
			"search": {
				"paths": ["/tmp/migfiletest08-45-21.93044607"],
				"md5": ["8d3a7afb7e59693b383d52396243a5b8"],
				"sha1": ["d82bc1145d471714b056940b268032f9ab0df2ae"],
				"sha2": ["3b495fae5bae9751ea4706c29e992002ba277bce30bd83a827b01ba977eabc2f"],
				"sha3": ["fdb23afa808c265284c3199013e4ded9704eebf54ffdc1f016dacc12"],
				"options": {
					"maxdepth": 0,
					"matchall": true,
					"macroal": false,
					"mismatch": ["md5", "sha1", "sha2", "sha3"],
					"matchlimit": 0
				}
			},
			"fileinfo": {
				"size": 0,
				"mode": "",
				"lastmodified": ""
			}
		}]
	},
	"statistics": {
		"filescount": 243,
		"openfailed": 0,
		"totalhits": 73,
		"exectime": "1.03271ms"
	},
	"errors": null
}
```
Where `"file": "",` indicates that nothing was found, but the test expected to find testfile0 and testfile1 because it `mismatch`es on the hashes of testfile2, and thus should find all other files.
I need to dig through your patch a big more to understand why it's failing. This should give you a clue to do the same.

jvehent:
Try rebasing against master, I merged a change in #159 that should fix the travis-ci issue.

jvehent:
@ameihm0912 : I'd like you to take a look at this change to make sure it doesn't break anything in scribe or mig-runner.

ameihm0912:
I don't see anything here that will cause problems in mig-runner or scribe
Do we want to also update sample actions in https://github.com/mozilla/mig/tree/master/actions based on these changes?

jvehent:
Ha, good call. @therealkbhat : could you update the examples, and maybe do a pass on the doc to make sure all references to the old sha* flags are updated?

therealkbhat:
Pushed doc and example updates.

jvehent:
This is good to go. I'm only going to delay merging it a bit because it breaks compatibility with existing agents (command line will generate `sha2` or `sha3` parameters that are unknown to older agents). We need to merge this right before redeploying a new version of the agent.

But it's safe to consider it ready to go. Thanks a lot @therealkbhat for the patch! :trophy:

jvehent:
@therealkbhat : could you resolve the conflicts on the branch?
We have an agent deployment scheduled for Jan.25th, so I'd like to merge this around the 20th.

therealkbhat:
@jvehent all systems green. :)

PR #160, opened 2015-12-22
[medium] Decompress files before content inspect, resolves #108

### LINKED ISSUE #108: [mig modules] automatically decompress files on the fly for content inspect in `file` module (Bugzilla #1116681)
Migrated from https://bugzilla.mozilla.org/show_bug.cgi?id=1116681

On 2014-12-30 19:50:12 -0800, Julien Vehent [:ulfr] wrote:

> The File module is able to inspect the content of files, but does not support automated decompression. This feature would be useful for scanning log files that are often compressed when rotated. It could also be used to search and inspect database dumps, or various suspicious files.
> 
> The performance impact of inspecting compressed file is an important aspect of this bug. MIG Agent must remain lightweight and not extravagantly consume CPU when scanning the file system.

therealkbhat:
Work in progress.

`decompressFile` returns a file descriptor which gets passed to `checkContent` and `checkHash` from `evaluateFile`. Note that I haven't removed `file string` from those calls; it seemed less invasive to pass both the string as well as the file descriptor, especially since the string is used extensively in calls to `debugprint` and `storematch`.

Does leaving the fd open across multiple function calls (`checkContent`, all the `checkHash`s) lead to any problems?

I haven't gotten to tests or doc changes yet; thought I'd get feedback on this approach first. I also still need to look at dealing with multiple files in a single zip.

### DIFF
--- a/modules/file/file.go
+++ b/modules/file/file.go
@@ -7,7 +7,7 @@
 /* The file module provides functions to scan a file system. It can look into files
 using regexes. It can search files by name. It can match hashes in md5, sha1,
 sha256, sha384, sha512, sha3_224, sha3_256, sha3_384 and sha3_512.
-The filesystem can be searches using pattern, as described in the Parameters
+The filesystem can be searched using patterns, as described in the Parameters
 documentation at http://mig.mozilla.org/doc/module_file.html .
 */
 package file /* import "mig.ninja/mig/modules/file" */
@@ -33,9 +33,12 @@ import (
 
 	"golang.org/x/crypto/sha3"
 	"mig.ninja/mig/modules"
+
+	"compress/gzip"
 )
 
 var debug bool = false
+var tryDecompress bool = false
 
 func debugprint(format string, a ...interface{}) {
 	if debug {
@@ -98,6 +101,7 @@ type options struct {
 	MatchLimit   float64  `json:"matchlimit"`
 	Debug        string   `json:"debug,omitempty"`
 	ReturnSHA256 bool     `json:"returnsha256,omitempty"`
+	Decompress   bool     `json:"decompress,omitempty"`
 }
 
 type checkType uint64
@@ -513,6 +517,11 @@ func (r *run) ValidateParameters() (err error) {
 			if err != nil {
 				return
 			}
+		}
+		if s.Options.Decompress {
+			tryDecompress = true
+		} else {
+			tryDecompress = false
 		}
 	}
 	return
@@ -934,6 +943,57 @@ func followSymLink(link string) (mode os.FileMode, path string, err error) {
 	return
 }
 
+type fileEntry struct {
+	filename string
+	fd       *os.File
+	compRdr  io.Reader
+}
+
+func (f *fileEntry) Close() {
+	f.fd.Close()
+}
+
+// getReader returns an appropriate reader for the file being checked.
+// This is done by checking whether the file is compressed or not
+func (f *fileEntry) getReader() io.Reader {
+	var (
+		err error
+	)
+	defer func() {
+		if e := recover(); e != nil {
+			err = fmt.Errorf("getReader() -> %v", err)
+		}
+	}()
+	f.fd, err = os.Open(f.filename)
+	if err != nil {
+		stats.Openfailed++
+		panic(err)
+	}
+	if tryDecompress != true {
+		return f.fd
+	}
+	magic := make([]byte, 2)
+	n, err := f.fd.Read(magic)
+	if err != nil {
+		panic(err)
+	}
+	if n != 2 {
+		return f.fd
+	}
+	_, err = f.fd.Seek(0, 0)
+	if err != nil {
+		panic(err)
+	}
+	if magic[0] == 0x1f && magic[1] == 0x8b {
+		f.compRdr, err = gzip.NewReader(f.fd)
+		if err != nil {
+			panic(err)
+		}
+		return f.compRdr
+	}
+	return f.fd
+}
+
 // evaluateFile takes a single file and applies searches to it
 func (r *run) evaluateFile(file string) (err error) {
 	var activeSearches []string
@@ -990,16 +1050,17 @@ func (r *run) evaluateFile(file string) (err error) {
 	// Second pass: Enter all content & hash checks across all searches.
 	// Only perform the searches that are active.
 	// Optimize to only read a file once per check type
-	r.checkContent(file)
-	r.checkHash(file, checkMD5)
-	r.checkHash(file, checkSHA1)
-	r.checkHash(file, checkSHA256)
-	r.checkHash(file, checkSHA384)
-	r.checkHash(file, checkSHA512)
-	r.checkHash(file, checkSHA3_224)
-	r.checkHash(file, checkSHA3_256)
-	r.checkHash(file, checkSHA3_384)
-	r.checkHash(file, checkSHA3_512)
+	f := fileEntry{filename: file}
+	r.checkContent(f)
+	r.checkHash(f, checkMD5)
+	r.checkHash(f, checkSHA1)
+	r.checkHash(f, checkSHA256)
+	r.checkHash(f, checkSHA384)
+	r.checkHash(f, checkSHA512)
+	r.checkHash(f, checkSHA3_224)
+	r.checkHash(f, checkSHA3_256)
+	r.checkHash(f, checkSHA3_384)
+	r.checkHash(f, checkSHA3_512)
 	return
 }
 
@@ -1151,10 +1212,9 @@ func (s search) checkMtime(file string, fi os.FileInfo) (matchedall bool) {
 	return
 }
 
-func (r *run) checkContent(file string) {
+func (r *run) checkContent(f fileEntry) {
 	var (
 		err error
-		fd  *os.File
 	)
 	defer func() {
 		if e := recover(); e != nil {
@@ -1184,14 +1244,10 @@ func (r *run) checkContent(file string) {
 	if !continuereadingfile {
 		return
 	}
-	fd, err = os.Open(file)
-	if err != nil {
-		stats.Openfailed++
-		panic(err)
-	}
-	defer fd.Close()
 	// iterate over the file content
-	scanner := bufio.NewScanner(fd)
+	reader := f.getReader()
+	defer f.Close()
+	scanner := bufio.NewScanner(reader)
 	for scanner.Scan() {
 		if err := scanner.Err(); err != nil {
 			panic(err)
@@ -1246,7 +1302,7 @@ func (r *run) checkContent(file string) {
 							debugprint("checkContent: [pass] must match any line and current line matched. regex='%s', line='%s'\n",
 								c.value, scanner.Text())
 							if c.wantThis(true) {
-								c.storeMatch(file)
+								c.storeMatch(f.filename)
 							}
 						}
 					} else {
@@ -1255,7 +1311,7 @@ func (r *run) checkContent(file string) {
 								c.value, scanner.Text())
 							macroalstatus[label] = false
 							if c.wantThis(true) {
-								c.storeMatch(file)
+								c.storeMatch(f.filename)
 							}
 						} else {
 							debugprint("checkContent: [fail] must not match at least one line but current line matched. regex='%s', line='%s'\n",
@@ -1280,7 +1336,7 @@ func (r *run) checkContent(file string) {
 								c.value, scanner.Text())
 							macroalstatus[label] = false
 							if c.wantThis(false) {
-								c.storeMatch(file)
+								c.storeMatch(f.filename)
 							}
 						} else {
 							debugprint("checkContent: [fail] much match any line and current line didn't match. regex='%s', line='%s'\n",
@@ -1309,7 +1365,7 @@ func (r *run) checkContent(file string) {
 		// 1. if MACROAL is set and the search succeeded, store the file in each content check
 		if search.Options.Macroal && macroalstatus[label] {
 			debugprint("checkContent: macroal is set and search label '%s' passed on file '%s'\n",
-				label, file)
+				label, f.filename)
 			// we match all content regexes on all lines of the file,
 			// as requested via the Macroal flag
 			// now store the filename in all checks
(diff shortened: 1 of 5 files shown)


therealkbhat:
I fixed the other errors (will push all fixes together later), but I'm wondering what to do with the fact that `getHash` is called in `buildResults` as well. One solution is to pass the file descriptor to `buildResults` in a similar fashion as done with `checkContent` and `checkHash`, but it requires another call to `decompressFile` and seems rather clunky.

ameihm0912:
I think this looks like a good start; it doesn't build for me but I believe you mentioned there were some changes still coming.

What if rather than pass a filename and descriptor around, we create a new abstract file struct that would include the filename, and just use that. So something like

type fileInfo struct {
        filename string
}

func (fileInfo *f) getReader() (io.Reader, error) {
        ...
}

In getReader() we could have the code you are using to detect the compression header, and return the correct Reader for use in the various content matching and hash functions. Then we just replace all the various Open() calls with getReader on the fileInfo type, and defer a call to close it.

This seems like it might be a bit cleaner and would eliminate the need to keep the open descriptor around, I'm not sure if we will run into issues with that. Thoughts?

therealkbhat:
Sounds good. Alternatively, it could be implemented as:
`func getScanner(file string) (*bufio.Scanner, err) {
...
}`
and we could replace the `scanner := bufio.NewScanner(fd)` line in `checkContent` with the call to `getScanner`.

There is still a problem in dealing with `getHash`, however. That function operates directly on `fd` (`fd.ReadAt(buf, offset)`). How do we modify that?

By the way, the `fileinfo` struct already exists.

ameihm0912:
How about the get reader function returning a section reader in all cases, which would support the call to ReadAt for the hash function in addition to creation of a scanner

therealkbhat:
That would work.
How do I go from `*gzip.Reader` to `*io.SectionReader`, in the getReader function? The argument to `SectionReader` is of type `io.ReaderAt`, which `*gzip.Reader` does not implement.

ameihm0912:
Was thinking something like this, here is a rough example:

https://gist.github.com/ameihm0912/42f5b8516b2d71a1297b

therealkbhat:
Can we assume that multiple files would normally be put in a tar structure before gzipping? gzip itself doesn't know anything about the file structure that it's compressing.

What do you think should be the next step here?

jvehent:
Do we care about files? Can't we just treat the content of the archive as a binary object and apply content regexes on it?

therealkbhat:
Ah, right. That makes sense.

Shall I move ahead with tests/docs? I also want to add support for the other Go standard compression libraries, although gzip is probably the most common of the lot.

modules/file/file.go
```diff
@@ -103,6 +105,7 @@ type options struct {
 	MatchLimit   float64  `json:"matchlimit"`
 	Debug        string   `json:"debug,omitempty"`
 	ReturnSHA256 bool     `json:"returnsha256,omitempty"`
+	Decompress   bool   `json:"decompress,omitempty"`
```
jvehent: You should `go fmt` your code before sending PRs. If you're a vim user, I strongly recommend you use [vim-go](https://github.com/fatih/vim-go) which does all that for you automatically.

therealkbhat: Oops, I'll fix that. Thanks!

modules/file/file.go
```diff
@@ -988,6 +991,49 @@ func followSymLink(link string) (mode os.FileMode, path string, err error) {
...
+	var err error
+	if f.checkCompressed() {
+		return f.compRdr
+	}
+	f.fd, err = os.Open(f.filename)
+	if err != nil {
+		panic(err)
+	}
+	return f.fd
+}
```
jvehent: I think the logic for the two functions above needs to be adjusted a bit:
1. in `checkCompressed`, if the file is not compressed, the descriptor is not closed, but another one is opened in `getReader`. It would be better to just reuse the first one.
2. I'm not sure we get much by splitting up `checkCompressed` and `getReader`, especially because `checkCompressed` does a lot more than `checking`. I would merge this two functions together and avoid the file descriptor dance, it's only 40 lines of code anyway.
3. don't `panic` without having an explicit `recover` in that function. In this case, `checkContent` has a recover and would catch the panic, but in MIG it is bad practice to recover outside of the function that panics. If your function needs to return an error, make it so:
```go
func myfunc() (my stuff, err error) {
    defer func() {
        if e:= recover(); e != nil {
            err = fmt.Errorf("myfunc() -> %v", err)
        }
    }()
    panic(err)
}
```

modules/file/file.go
```diff
@@ -1238,14 +1284,9 @@ func (r *run) checkContent(file string) {
 	if !continuereadingfile {
 		return
 	}
-	fd, err = os.Open(file)
-	if err != nil {
-		stats.Openfailed++
```
jvehent: You removed this counter, but has it been readded somewhere else?

modules/file/file.go
```diff
@@ -1465,19 +1506,14 @@ func (r *run) checkHash(file string, hashtype checkType) {
...
+func getHash(f fileEntry, hashType checkType) (hexhash string, err error) {
+	reader := f.getReader()
 	defer func() {
 		if e := recover(); e != nil {
 			err = fmt.Errorf("getHash() -> %v", e)
 		}
 	}()
-	fd, err := os.Open(file)
-	if err != nil {
-		stats.Openfailed++
```
jvehent: Same comment about the removed counter.

modules/file/file.go
```diff
@@ -1465,19 +1506,14 @@ func (r *run) checkHash(file string, hashtype checkType) {
 }
 
 // getHash calculates the hash of a file.
-func getHash(file string, hashType checkType) (hexhash string, err error) {
+func getHash(f fileEntry, hashType checkType) (hexhash string, err error) {
+	reader := f.getReader()
```
jvehent: should move this after the recover block

modules/file/file.go
```diff
@@ -1536,6 +1570,12 @@ type fileinfo struct {
 	SHA256 string  `json:"sha256,omitempty"`
 }
 
+type fileEntry struct {
+	filename string
+	fd       *os.File
+	compRdr  io.Reader
+}
```
jvehent: I'd move this right before the `getReader` function, since they work together.

modules/file/paramscreator.go
```diff
@@ -80,6 +80,8 @@ Options
 			  is reached.
 %sreturnsha256		- include sha256 hash for matched files.
 			  ex: %sreturnsha256
+%sdecompress	- decompress file before inspection
```
jvehent: align

jvehent:
The logic is sound. This needs a little more work, tons of tests and some documentation, but it's definitely on the right path. :+1:

modules/file/file.go
```diff
@@ -1238,14 +1284,9 @@ func (r *run) checkContent(file string) {
...
 	}
-	fd, err = os.Open(file)
-	if err != nil {
-		stats.Openfailed++
-		panic(err)
-	}
-	defer fd.Close()
 	// iterate over the file content
-	scanner := bufio.NewScanner(fd)
+	reader := f.getReader()
```
jvehent: Who's responsible for closing that reader?

therealkbhat:
We're opening the file to be inspected *before* running through each search, right? However, we're using the decompress flag to set which reader we return, which is specific to a particular search. How do I resolve this and decompress files only if `search.Options.Decompress` is set?

This also got me thinking: is there a situation where we *don't* want to unzip a zipped file? If not, we could just have decompression as part of the normal sequence of operations.

jvehent:
> How do I resolve this and decompress files only if search.Options.Decompress is set?

Let's make this option a special case, and clearly state in the doc that if it's set for one search, all searches will use it. People can always trigger multiple runs of the file module to work around this if needed.

> This also got me thinking: is there a situation where we don't want to unzip a zipped file? If not, we could just have decompression as part of the normal sequence of operations.

Yes. Quite often actually. There are searches where we know the target file is not gzipped, and thus do not care to gzip anything as this would take cpu time unnecessarily.

jvehent:
I see test :+1: @therealkbhat is this ready for a final round of reviews?

therealkbhat:
@jvehent hehe, not quite. Let me add in some doc changes. :)

jvehent:
Make sure to add specific decompression tests to the compressed files. This might require to store a new content test to check the string on the decompressed file, rather than on the compressed flie. Same for hashes.

therealkbhat:
As I see it, I need to be able to walk through the set of testcases (in `func TestCompressedHashes`, say) and check if the file is zipped. If so, decompress, and perform the checks on the content.
I'm having some trouble with this...
* Is there a way I can perform these checks without saving the files to disk?
* Won't this involve including extra fields on all testcases? Or is it sufficient to have these fields (corresponding to content regex and hash values of the decompressed file) only for the files we know to be zipped?

jvehent:
> Is there a way I can perform these checks without saving the files to disk?

No. First the files are written to disk, then the tests search for them.

> Won't this involve including extra fields on all testcases?

Yes. I would suggest to add `decompressedcontent` and `decompressedhash` to the tests data of compressed files only. The other files don't need them.

> Or is it sufficient to have these fields (corresponding to content regex and hash values of the decompressed file) only for the files we know to be zipped?

No, because the current tests need to check that the `content` and `hash` of compressed files are found. That way, we can test both current behavior and new, decompressed, behavior.

therealkbhat:
Alright. So, the right thing to do here (for both content and hash checks) is to create a new function that will:
* check whether the test case corresponds to a zipped file, either by checking the header or the first two bytes of the `data` field.
* if it does, then decompress and write this content to another temp file, and set the `expectedfiles` variable accordingly. The remainder is effectively the same as the existing sequence of instructions.
* if not, skip it and move on.

Is this right? And if it is, is there a way to make use of the `getReader` function in `file.go` to cut down on the repetition of gzip file reading code?

therealkbhat:
@jvehent done, but not perfect.

testfile5 is a corrupted gzip file, and because of this, we never complete the content and hash checks (I've exited with an error if it fails to gunzip a file with the gzip header). I wasn't sure how to handle this, so currently I'm doing a check for `tp.name == "testfile5"` and proceeding. Very clunky.

Ideas?

modules/file/file_test.go
```diff
@@ -772,4 +863,92 @@ some other other text`),
...
+		sha1:                `bd1f24d8cbb000bbf7bcd618c2aec73280388721`,
+		sha256:              `bb4e449df74edae0292d60d2733a3b1801d90ae23560484b1e04fb52f111a14f`,
+		sha384:              `7f39be66900ee6a00796c02fea06ca8c5b787355d02ce37abcf5dc0685e457f34faf9559c4a9619e9bff1a6724c1d548`,
+		sha512:              `eacc268db1c9722e238e002db3f902a885ed2998fc4dc95b6f6582f3f73d0ad3e6cda8ef6f221657441eb347f7f1fff45c7483c1e97c1067b52500ab1eb17608`,
+		sha3_224:            `433b84f162d1b00481e6da022c5738fb4d04c3bb4317f73266746dd1`,
+		sha3_256:            `6ec0f8c51330796eb3fc6418f04bcd86ab89a195f6b8f6668f06d75d8fab2776`,
+		sha3_384:            `9fc6e87537293f29608e5eb8c9e6e4af73136b4e03165962dfc590cb8189625cd4c17c3b05b9aef1ab76982c80457eb0`,
+		sha3_512:            `2d7da22c6ff126df89b1393a6c500d7670259dbdd059164646b8ddae04704273a0ff3c1dd6d8562d6452d42fd85624b51720e403491b2e07aaefbfd4e5ed73fa`,
+		decompressedcontent: `^--- header for zipped file ---$`,
+		decompressedhash:    `5bb23d3b1eaaddc6e108e3fb0ee80a61`,
```
jvehent: for clarity, this should be `decompressedmd5`. Same in testfile4.

modules/file/file_test.go
```diff
@@ -102,6 +102,47 @@ func TestContentSearch(t *testing.T) {
...
 }
 
+func TestDecompressedContentSearch(t *testing.T) {
+	for _, tp := range TESTDATA {
+		var (
+			r run
+			s search
+		)
+		var expectedfiles = []string{""}
+		if tp.name != "testfile5" {
```
jvehent: I would prefer to do the opposite here, and have a special case that resets `expectedfiles` to an empty array if `tp.name == "testfile5"`.
A comment explaining that testfile5 is a corrupt gzip that should fail to decompress would be nice too.

jvehent:
I think the exception for `testfile5` is fine, if there's a comment both in the tests and above its testdata entry.

Would it be worth testing the error message returned when searching for `testfile5` with `-decompress` only, to verify that the file module return the right error?

Other than that, r+. I think this is very close to being merged.

jvehent:
Could you also resolve the conflicts created by merging #158 ?

therealkbhat:
@jvehent pushed changes.

> Would it be worth testing the error message returned when searching for `testfile5` with `-decompress` only, to verify that the file module return the right error?

I'm not sure... we don't really gain much by doing so. Ultimately, if the file module *doesn't* return the right error there, we're just going to raise another one, right?

modules/file/file.go
jvehent: If this comment is addressed, please reference the commit that addresses it.

modules/file/file.go
jvehent: If this comment is addressed, please reference the commit that addresses it.

modules/file/file.go
jvehent: If this comment is addressed, please reference the commit that addresses it.

modules/file/file.go
jvehent: If this comment is addressed, please reference the commit that addresses it.

modules/file/file.go
jvehent: nevermind, just noticed it's closed below :+1:

modules/file/file.go
jvehent: nvm, github left the comment even though you moved the code. All good!

modules/file/file.go
jvehent: found it in getReader. all good.

PR #161, opened 2015-12-24
[medium] Support fetch of instance meta-data for agents running in AWS

ameihm0912:
This change adds support for discovery of specific elements of instance metadata by agents running in AWS. Fetched metadata is included in the agents environment.

### DIFF
--- a/agent.go
+++ b/agent.go
@@ -36,14 +36,18 @@ type Agent struct {
 
 // AgentEnv stores basic information of the endpoint
 type AgentEnv struct {
-	Init      string   `json:"init,omitempty"`
-	Ident     string   `json:"ident,omitempty"`
-	OS        string   `json:"os,omitempty"`
-	Arch      string   `json:"arch,omitempty"`
-	IsProxied bool     `json:"isproxied"`
-	Proxy     string   `json:"proxy,omitempty"`
-	Addresses []string `json:"addresses,omitempty"`
-	PublicIP  string   `json:"publicip,omitempty"`
+	Init            string   `json:"init,omitempty"`
+	Ident           string   `json:"ident,omitempty"`
+	OS              string   `json:"os,omitempty"`
+	Arch            string   `json:"arch,omitempty"`
+	IsProxied       bool     `json:"isproxied"`
+	Proxy           string   `json:"proxy,omitempty"`
+	Addresses       []string `json:"addresses,omitempty"`
+	PublicIP        string   `json:"publicip,omitempty"`
+	AWSInstanceID   string   `json:"aws_instanceid,omitempty"`
+	AWSLocalIPV4    string   `json:"aws_localipv4,omitempty"`
+	AWSAMIID        string   `json:"aws_amiid,omitempty"`
+	AWSInstanceType string   `json:"aws_instancetype,omitempty"`
 }
 
 type AgentsStats struct {
--- a/agent.go
+++ b/agent.go
@@ -49,10 +49,10 @@ type AgentEnv struct {
 
 // Stores AWS specific agent environment values
 type AgentEnvAWS struct {
-	AWSInstanceID   string `json:"aws_instanceid,omitempty"`
-	AWSLocalIPV4    string `json:"aws_localipv4,omitempty"`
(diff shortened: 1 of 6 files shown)


agent.go
```diff
@@ -36,14 +36,18 @@ type Agent struct {
...
+	OS              string   `json:"os,omitempty"`
+	Arch            string   `json:"arch,omitempty"`
+	IsProxied       bool     `json:"isproxied"`
+	Proxy           string   `json:"proxy,omitempty"`
+	Addresses       []string `json:"addresses,omitempty"`
+	PublicIP        string   `json:"publicip,omitempty"`
+	AWSInstanceID   string   `json:"aws_instanceid,omitempty"`
+	AWSLocalIPV4    string   `json:"aws_localipv4,omitempty"`
+	AWSAMIID        string   `json:"aws_amiid,omitempty"`
+	AWSInstanceType string   `json:"aws_instancetype,omitempty"`
```
jvehent: I'd prefer having a separate AWS struct that contains a bunch of instance information, as opposed to adding them directly into the Env struct. We could always query them using `agents.environment->'aws'->>'instanceid'`.

agent.go
```diff
@@ -36,14 +36,18 @@ type Agent struct {
...
+	Init            string   `json:"init,omitempty"`
+	Ident           string   `json:"ident,omitempty"`
+	OS              string   `json:"os,omitempty"`
+	Arch            string   `json:"arch,omitempty"`
+	IsProxied       bool     `json:"isproxied"`
+	Proxy           string   `json:"proxy,omitempty"`
+	Addresses       []string `json:"addresses,omitempty"`
+	PublicIP        string   `json:"publicip,omitempty"`
+	AWSInstanceID   string   `json:"aws_instanceid,omitempty"`
+	AWSLocalIPV4    string   `json:"aws_localipv4,omitempty"`
```
jvehent: Is this different from the content of `addresses` above? If yes, can it be merged into the addresses?

mig-agent/aws.go
```diff
@@ -0,0 +1,130 @@
...
+		Dial: (&net.Dialer{Timeout: time.Second}).Dial,
+	}
+	client := &http.Client{Transport: tr}
+	resp, err := client.Get(awsMetaURL() + endpoint)
+	if err != nil {
+		return
+	}
+	defer resp.Body.Close()
+	if resp.StatusCode != http.StatusOK {
+		err = fmt.Errorf("invalid HTTP response code returned by metadata service")
```
jvehent: maybe log the response code here, for debugging?

jvehent:
Looks good to me. I was looking for a way to just retrieve a whole JSON from the meta-data in one call, but there doesn't seem to be one. Failing that, you approach is good, I'd just change the struct to put everything into an `AWS` struct.

/me likes :+1:

ameihm0912:
Yeah, I also looked for a similar interface to just get all the meta-data, but there doesn't seem to be anything. It would also be nice if there was a way to obtain instance tags through the same meta-data calls, but this doesn't seem to be possible either.

With the instance-id however, we should be able to cross reference the instance with data obtained from the AWS API that describes the instances, so we can add tags if we need to. This could be done post-request (e.g. as part of a mig-runner processing job).

The suggested changes make sense, I'll take a look at updating the PR.

agent.go
```diff
@@ -36,14 +36,23 @@ type Agent struct {
...
+	PublicIP  string      `json:"publicip,omitempty"`
+	AWS       AgentEnvAWS `json:"aws,omitempty"`
+}
+
+// Stores AWS specific agent environment values
+type AgentEnvAWS struct {
+	AWSInstanceID   string `json:"aws_instanceid,omitempty"`
+	AWSLocalIPV4    string `json:"aws_localipv4,omitempty"`
+	AWSAMIID        string `json:"aws_amiid,omitempty"`
+	AWSInstanceType string `json:"aws_instancetype,omitempty"`
```
jvehent: I'd drop the `aws_` prefixes in the variables and json, we're already under the `aws` key.

jvehent:
Just one nitpick and it's ready to merge.

PR #162, opened 2015-12-29
[medium] Add a runner plugin to process compliance items

ameihm0912:
Adds a plugin for use with mig-runner which replaces the existing compliance worker process


jvehent:
I would suggest removing the compliance worker and migrating the documentation of it into the mig-runner doc. Maintaining a single compliance generation tool is probably enough, and mig-runner provides a better approach in doing so anyway.

r+ on the code.

ameihm0912:
Works for me, I've removed the old code and migrated the documentation.

jvehent:
Looks good to me. r+

PR #164, opened 2015-12-30
[major] sandbox implementation

tudalex:


modules/file/file.go
```diff
@@ -31,6 +31,8 @@ import (
 	"strings"
 	"time"
 
+	"github.com/seccomp/libseccomp-golang"
```
gdestuynder: libseccomp-golang should probably only be loaded in the Linux environment

modules/agentdestroy/agentdestroy.go
```diff
@@ -14,6 +14,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"github.com/kardianos/osext"
+	"github.com/tudalex/seccomp-sandbox"
```
gdestuynder: The new repo is github.com/mozilla/mig-sandbox which should eventually contain sandbox code for other platforms as well (non-linux will just be empty for now)

modules/file/file.go
```diff
@@ -44,14 +46,52 @@ func debugprint(format string, a ...interface{}) {
...
+			sandbox.FilterOperation{
+				FilterOn: []string{
+					"pread64",
+					"lstat",
+					"close",
+					"openat",
+					"stat",
+					"readlinkat",
+					"write",
+					"mmap",  // GO
```
gdestuynder: nitpick: would clarify "//GO" into "required by Go runtime" or something like that

modules/file/file.go
```diff
@@ -44,14 +46,52 @@ func debugprint(format string, a ...interface{}) {
...
+	m := new(module)
+	sandbox := sandbox.SandboxProfile{
+		DefaultPolicy: seccomp.ActTrap,
+		Filters: []sandbox.FilterOperation{
+			sandbox.FilterOperation{
+				FilterOn: []string{
+					"pread64",
+					"lstat",
+					"close",
+					"openat",
```
gdestuynder: openat(): Would flag this as unsafe (ie we should try to remove it or filter it better in the future)

modules/memory/memory.go
```diff
@@ -28,14 +30,50 @@ import (
...
+
 func init() {
-	modules.Register("memory", new(module))
+	m := new(module)
+	sandbox := sandbox.SandboxProfile{
+		DefaultPolicy: seccomp.ActTrap,
+		Filters: []sandbox.FilterOperation{
+			sandbox.FilterOperation{
+				FilterOn: []string{
+					"openat",
```
gdestuynder: Same as for file module. in this case i suspect it has to open a bunch of things though, so i wonder if we'll have to filter it later or something like that.

modules/netstat/netstat.go
```diff
@@ -22,14 +24,32 @@ import (
...
+}
+
 func init() {
-	modules.Register("netstat", new(module))
+	m := new(module)
+	sandbox := sandbox.SandboxProfile{
+		DefaultPolicy: seccomp.ActTrap,
+		Filters: []sandbox.FilterOperation{
+			sandbox.FilterOperation{
+				FilterOn: []string{"write", "read", "close",
```
gdestuynder: nitpick: this syntax differs from the previous modules, should probably use the same syntax

modules/netstat/netstat.go
```diff
@@ -22,14 +24,32 @@ import (
...
+					"futex", "openat", "socket", "exit_group",
```
gdestuynder: write()+openat() same comment as for file module, mark unsafe

modules/ping/ping.go
```diff
@@ -26,14 +28,56 @@ import (
...
+					"epoll_ctl",
+					"select",
+					"setsockopt",
+					"close",
+					"socket",
+					"connect",
+					"futex",
+					"getsockname",
+					"getpeername",
+					"write",
```
gdestuynder: another write()+openat()

modules/timedrift/timedrift.go
```diff
@@ -25,14 +27,58 @@ import (
...
-	modules.Register("timedrift", new(module))
+	m := new(module)
+	sandbox := sandbox.SandboxProfile{
+		DefaultPolicy: seccomp.ActTrap,
+		Filters: []sandbox.FilterOperation{
+			sandbox.FilterOperation{
+				FilterOn: []string{
+					"select",
+					"futex",
+					"write",
```
gdestuynder: write() + openat()

vendor/github.com/tudalex/seccomp-sandbox/sandbox.go
```diff
@@ -0,0 +1,98 @@
+package sandbox
+
+/*
```
gdestuynder: s/sandbox.go/sandbox_linux.go for cross platform compat

vendor/github.com/tudalex/seccomp-sandbox/sandbox.go
```diff
@@ -0,0 +1,98 @@
...
..._at", "clock_adjtime", "syncfs", "sendmmsg", "setns", "getcpu", "process_vm_readv", "process_vm_writev", "kcmp", "finit_module", "sched_setattr", "sched_getattr", "renameat2", "seccomp", "getrandom", "memfd_create", "kexec_file_load", "bpf", "execveat", "unkown syscall 323", "unkown syscall 324", "unkown syscall 325", "unkown syscall 326", "unkown syscall 327", "unkown syscall 328", "unkown syscall 329", "unkown syscall 330", "unkown syscall 331", "unkown syscall 332", "unkown syscall 333", "unkown syscall 334", "unkown syscall 335", "unkown syscall 336", "unkown syscall 337", "unkown syscall 338", "unkown syscall 339", "unkown syscall 340", "unkown syscall 341", "unkown syscall 342", "unkown syscall 343", "unkown syscall 344", "unkown syscall 345", "unkown syscall 346", "unkown syscall 347", "unkown syscall 348", "unkown syscall 349", "unkown syscall 350", "unkown syscall 351", "unkown syscall 352", "unkown syscall 353", "unkown syscall 354", "unkown syscall 355", "unkown syscall 356", "unkown syscall 357", "unkown syscall 358", "unkown syscall 359", "unkown syscall 360", "unkown syscall 361", "unkown syscall 362", "unkown syscall 363", "unkown syscall 364", "unkown syscall 365"};
```
gdestuynder: add as comment where the mapping is from (ie kernel headers path for ex)

modules/timedrift/timedrift.go
tudalex: I think you need write in order to write the results back on the pipe and also for our sandboxing code to print the error message back. We'll try to see if we can filter it to be only to fd 1 and 2.

vendor/github.com/tudalex/seccomp-sandbox/sandbox.go
```diff
@@ -0,0 +1,98 @@
...
+					log.Fatal("Error adding rule to restrict syscall: %s\n", err)
+				} else {
+					log.Printf("Added rule to restrict syscall %s\n", callName)
+				}
+			}
+		}
+	}
+	filter.SetTsync(true)
+	filter.SetNoNewPrivsBit(true)
+	err = filter.Load()
```
gdestuynder: potentially, we could add code to detect if seccomp is supported here, for ex:
if it return errBadFilter then we have a filter error, log that the filter format appears incorrect/syntax error, fail and exit here (as in mig terminates)

Otherwise, check the syscall return, it should return unsupported or permission denied when seccomp is not supported, which we could log and continue, such as:

log.Printf("Failed to load seccomp filter, seccomp (filter mode) not supported by kernel: %s. Sandbox features disabled.", err)

See also: https://github.com/seccomp/libseccomp-golang/blob/master/seccomp.go#L592

modules/file/file.go
```diff
@@ -44,14 +46,52 @@ func debugprint(format string, a ...interface{}) {
...
+}
+
 func init() {
-	modules.Register("file", new(module))
+	m := new(module)
+	sandbox := sandbox.SandboxProfile{
+		DefaultPolicy: seccomp.ActTrap,
+		Filters: []sandbox.FilterOperation{
+			sandbox.FilterOperation{
+				FilterOn: []string{
```
gdestuynder: Note, could also use SetSyscallPriority for priority setting - not sure if it makes a difference when the filter is ordered from the point of view of libseccomp

vendor/github.com/tudalex/seccomp-sandbox/sandbox.go
ameihm0912: In addition to this it could be useful to return some sort of status indication from this function to the caller so the agent could decide if it should continue if the jail failed, potentially based on a compile-time flag or configuration setting.

vendor/github.com/tudalex/seccomp-sandbox/sandbox.go
ameihm0912: Yeah we'll need to figure out what we want to do for the other platforms here. We might need to have stub functions that do nothing for Windows and OSX so the build does not fail on these platforms when it's linked with this library. I imagine we could do that here, or if we want to keep this library seccomp focused, we could wrap the jail call in another function in mig (e.g., sandbox_linux.go would contain a function that calls this, and imports the library), _windows _osx would just be a noop

modules/timedrift/timedrift.go
gdestuynder: note if anything for all these, id just put a comment on why they're here / why we think its dangerous, so that it can be enhanced later

jvehent:
What's up guy? Can this be cleaned up to go through a round of reviews?

gdestuynder:
@tudalex  @teobaluta  @vladimirdiaconescu I keep the hoodies hostage until this is merged! ;-)

gdestuynder:
I've fixed the conflicts and re-organized the commits in a new PR to be mergeable as we've having troubles reaching the original participants. The new PR is at https://github.com/mozilla/mig/pull/230
Feel free to reach out though!

PR #172, opened 2016-01-20
[minor] Update description of hash inspection capabilities

therealkbhat:
Ultra minor.
I was going through docs and saw that this needed updating.


jvehent:
I would just say
`with md5 and sha1/2/3 hashes`

therealkbhat:
That does sound cleaner. :)

PR #174, opened 2016-01-22
[minor] Assorted doc fixes

therealkbhat:
Another ultra-minor one; consists primarily of spelling fixes and the odd rewording of a phrase.


ameihm0912:
r+

jvehent:
Thanks for the patch!
Two small fyi: 1) you can omit the HTML files from your PRs, I'll regenerate them later and 2) you can use commit prefix `[doc]` instead of `[minor]` when the commit only touches documentation.

therealkbhat:
@jvehent I'll keep that in mind, thanks.

PR #176, opened 2016-01-22
[medium] introduce optional action compression between client and agent

ameihm0912:
Introduces optional compression of actions between the client and the agents

Compression is specified during action creation and handled when processed by the agent

### DIFF
--- a/modules/modules_test.go
+++ b/modules/modules_test.go
@@ -7,6 +7,10 @@
 package modules /* import "mig.ninja/mig/modules" */
 
 import (
+	"bytes"
+	"compress/gzip"
+	"encoding/base64"
+	"encoding/json"
 	"io"
 	"os"
 	"strings"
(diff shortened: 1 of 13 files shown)


modules/modules_test.go
```diff
@@ -68,7 +72,29 @@ func TestMakeMessage(t *testing.T) {
 		t.Fatalf("Invalid module message class `parameters`")
 	}
 
-	raw, err = MakeMessage(MsgClassStop, nil)
+	// Test parameter decompression
```
jvehent: For clarity, I'd recommend making this its own test, maybe under `TestDecompressedAndMakeMessage` ?

jvehent:
That looks good.
Do you think there's a way to move the compression code from the console and cmdline into the base client package? Maybe as part of `PostAction` ? If not, I think your approach is sane.

Issue #100, opened 2015-10-02
[mig modules] create `account` module to lookup user activity on endpoints (Bugzilla #1037965)


jvehent:
Migrated from https://bugzilla.mozilla.org/show_bug.cgi?id=1037965
Assigned to: Julien Vehent [:ulfr]

On 2014-07-13 06:41:50 -0700, Julien Vehent [:ulfr] wrote:

> Provide a standard module that can verify the presence of a given account or group on a system. It should also be able to create and disable accounts, or compare a given hash string with the current password (to check when a password has not been properly changed).

On 2014-12-31 10:40:51 -0800, Julien Vehent [:ulfr] wrote:

> One interesting feature would be to parse /var/log/lastlog to check where a user has recently logged in.

PR #177, opened 2016-01-22
[medium] [module] Account module for looking up user activity

therealkbhat:
This is in reference to #100, and is still a work in progress.
So far, the "check user/group on system" aspect has been tackled with.
I thought I'd raise a PR early so that I can correct any mistakes and/or
incorrect assumptions sooner rather than later.

For ongoing progress (and as a note to self):
- [X] check if user/group is present on system
- [X] ~~create and disable accounts~~ (makes more sense to move this to a new module, see below)
- [X] ~~compare hash string with current password~~ (avoiding this due to violation of the "don't extract data" policy)
- [ ] check when a user last logged in -- this can be done by parsing `/var/log/auth.log` for `sshd` and `systemd-logind` entries
- [ ] documentation!

### DIFF
--- a/conf/available_modules.go
+++ b/conf/available_modules.go
@@ -6,6 +6,7 @@
 package main
 
 import (
+	_ "mig.ninja/mig/modules/account"
 	_ "mig.ninja/mig/modules/agentdestroy"
 	_ "mig.ninja/mig/modules/file"
 	_ "mig.ninja/mig/modules/memory"
--- a/modules/account/account.go
+++ b/modules/account/account.go
@@ -1,0 +1,215 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/.
+//
+// Contributor: Kishor Bhat kishorbhat@gmail.com [:kbhat]
+
+package account /* import "mig.ninja/mig/modules/account" */
+
+import (
+	"bufio"
+	"encoding/json"
+	"fmt"
+	"io"
+	"mig.ninja/mig/modules"
+	"os"
+	"regexp"
+	"strings"
+)
+
+var debug bool = false
+
+func debugprint(format string, a ...interface{}) {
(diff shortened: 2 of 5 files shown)


jvehent:
>  create and disable accounts

I wonder if this would be better as a separate module, so we can set specific ACLs on it, like requiring a double signature. The rest of the module is only reading data, so everyone could have access to it, but creating users is a sensitive operation.

>  compare hash string with current password

I'm not sure we want to pass hashed passwords through mig, this falls under the "dont return system data" category.

>  check when a user last logged in

This would be great if it included the SSH logs lookup.

therealkbhat:
> The rest of the module is only reading data, so everyone could have access to it, but creating users is a sensitive operation.

Makes sense; but doesn't it seem like overkill to have a module that *just* creates/disables users?

> this falls under the "dont return system data" category.

Ah yes, I forgot to take that into consideration.

> This would be great if it included the SSH logs lookup.

Sounds good to me!

jvehent:
> doesn't it seem like overkill to have a module that just creates/disables users?

Not really. Modules are cheap. It's better to have smaller ones with specific permissions that a few larges ones.

ameihm0912:
Can we not just use the regular bufio functions like ReadLine() here?

I think the approach makes sense. Only concern I have is it could be fairly cumbersome making sure we do all these checks and handle various cases (e.g., using /var/log/secure, other paths based on distribution). I suppose that is unavoidable though.

@jvehent this is really a case where some sort of generic lower-level log reading module would come in handy, so other modules do not need to be aware of the underlying logging architecture.

e.g., a module where we could say, "give me the newest 1000 log lines that match this criteria" without specifying a file name. from here a module like this could just make use of that facility. not sure if that is overkill or worth considering here. if we had something like that, that would also form the basis for doing more on-going monitoring and automating selection of what we want to look at in the agent.

jvehent:
I wonder how much of what you're describing is provided by systemd-journald. It's worth looking into.

ameihm0912:
Closing PR due to inactivity, please recreate if still a WIP

PR #179, opened 2016-01-29
[minor] strip unicode NULL prior to database insertion

ameihm0912:


jvehent:
Looks good to me. I guess we could theoretically do without the safety `unmarshal`, since \u0000 doesn't correspond to any json reserved character there no reason why converting them would break the structure, but at the same time it's not an expensive call to make, so might as well.

PR #183, opened 2016-02-04
Integration of mig loader

ameihm0912:
PR related to integration of MIG loader; note this is a work in progress but basic functionality is present. There are some missing elements yet to be added, and some cleanup.


doc/loader.rst
```diff
@@ -0,0 +1,81 @@
...
+---------
+As an example, the loader would operate as follows.
+
+The loader binary would be installed on a client system using a package in the
+same way the agent would be installed now. This package would also be
+responsible for adding an operation to the scheduling facility on the system
+(for example, cron, Windows Scheduler) to run `mig-loader` on a periodic basis
+on the system.
+
+When the loader is run by the scheduler, it will create a manifest based on
```
jvehent: I don't understand this: why does the scheduler run the loader?

doc/loader.rst
```diff
@@ -0,0 +1,81 @@
...
+what MIG components/configuration files are present on the system. It is likely
+this will be limited to the agent itself, and configuration files used by the
```
jvehent: is this use to detect the version of mig-agent currently installed on the system and report that back to the api?

doc/loader.rst
```diff
@@ -0,0 +1,81 @@
...
+Next, the loader will generate an agent environment structure. This will
+include the same information the agent typically sends as part of it's
+environment (e.g., the system hostname, operator identifier, others).
+
+Once the loader has this information, it will request the current manifest
+from the API by sending it's environment. The API will look at the environment,
+and determine the current manifest for use on this particular system. It is
+intended this information will be stored in the MIG database. Once the API
+has made this determination, it will respond to the loader with what the
+manifest should be.
```
jvehent: Does the api return a url to the manifest to be retrieved from a different location, or the manifest itself?

doc/loader.rst
```diff
@@ -0,0 +1,81 @@
...
+
+When the loader recieves this manifest, it will compare it with the manifest
+it generated using local information from the system. If there are mismatches,
+the loader will make a request to the API to fetch the files which are not
+current, stage the files on the file system, and eventually replace the files
+with the new versions.
```
jvehent: I'm wondering if we care to capture this information somehow. I'm thinking that a manifest mismatch may indicate the agent binary was modified locally, which is an anomaly.

doc/loader.rst
```diff
@@ -0,0 +1,81 @@
...
+
+In cases where there are changes, the loader will need to look after initiating
+the new agent code. This would for example involve spawning a new mig-agent
+process, which in turn will result in the old agent being killed. In cases
+where it is a new system (no current mig-agent) this would also cause the agent
+to self-install itself in the service manager for the OS.
+
+Loader updates
+--------------
+It is anticipated the loader will requires updates far less likely than the
```
jvehent: aka. "never" :)

jvehent:
Since you're moving environment detection to its own package, would it be worth to make it a proper agent module? That would let us query environments in realtime too.

mig-api/api.go
```diff
@@ -70,6 +70,12 @@ func main() {
 	// unauthenticated endpoints
 	s.HandleFunc("/heartbeat", getHeartbeat).Methods("GET")
 	s.HandleFunc("/ip", getIP).Methods("GET")
+	// Loader manifest endpoints
+	s.HandleFunc("/manifest", authenticate(getManifest)).Methods("GET")
+	s.HandleFunc("/manifest/sign/", authenticate(signManifest)).Methods("POST")
+	s.HandleFunc("/manifest/status/", authenticate(statusManifest)).Methods("POST")
+	s.HandleFunc("/manifest/agent/", getAgentManifest).Methods("POST")
+	s.HandleFunc("/manifest/fetch/", getManifestFile).Methods("POST")
```
jvehent: if we're `GET`ing manifests, the methods should be `GET`, not `POST`. (POST is only used when some state on the server side is changed).

jvehent:
I'm having trouble reviewing because of how massive it is. I think I'll need to build it locally and test to understand the workflow better. It might be worth splitting up the PR into multiples to facilitate review, for example:
1. agent/load env detection
2. manifest handling in api/database
3. loader implementation

database/loader.go
```diff
@@ -0,0 +1,96 @@
...
+	_ "github.com/lib/pq"
+	"mig.ninja/mig"
+)
+
+// Return a loader entry ID given a loader key
+func (db *DB) GetLoaderEntryID(key string) (ret float64, err error) {
+	if key == "" {
+		return ret, fmt.Errorf("key cannot be empty")
+	}
+	rows, err := db.c.Query("SELECT id FROM loaders WHERE loaderkey=$1", key)
```
jvehent: If you expect one row, you can use `QueryRow` directly.

PR #186, opened 2016-02-23
WIP agent service refactoring

jvehent:
not functional, but for future ref

### DIFF
--- a/mig-agent/context.go
+++ b/mig-agent/context.go
@@ -13,33 +13,30 @@ import (
 	"crypto/x509"
 	"encoding/json"
 	"fmt"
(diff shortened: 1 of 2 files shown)


robmurtha:
@jvehent can you please rebase this against the latest?

jvehent:
@robmurtha No, I don't think we can rebase it without significant rewrite, unfortunately. This patch was a backup of work that was startup but never finished. :(

ameihm0912:
@jvehent can this one be closed? This is a ways out of date as the agent service code has been refactored quite a bit.

jvehent:
Totally, this is ancient.

PR #189, opened 2016-03-02
[minor/bug] Fix stand alone install script to accept go version > 1.5

arunk-s:


tools/standalone_install.sh
```diff
@@ -74,8 +74,7 @@ if [ ! $(which go) ]; then
     exit 1
 fi
 go_version=$(go version)
-# This expression will need to be updated for future releases.
-echo $go_version | grep -q -E 'go version go1\.5'
+echo $go_version | grep -E -q --regexp="go[1-9]\.[5-9]"
```
jvehent: this will break for version `[2-9]\.[0-4]` and doesn't allow go version > 1.9 (eg. 1.10)

Since we're now at version 1.6, and ubuntu supports 1.5, can we drop this test entirely? @ameihm0912 ?

arunk-s: @jvehent Sorry, my bad. I didn't think thoroughly on the other test cases. If decided to keep this test, I'll see other ways to implement this.

ameihm0912:
I'm not sure about other distributions, what if we just invert the logic here, make sure we get a valid result back from go version and it doesn't match <1.5 ? Since the only thing we are looking for here is making sure vendoring support exists.

jvehent:
something like: `echo $go_version | grep -E -q --regexp="go1\.[0-4]" && fail`

arunk-s:
or maybe just:
```bash
echo $go_version | grep -E -q --regexp="go1\.[0-4]"
if [ $? -ne 1 ]; then
```
This also means the same thing, right ?

jvehent:
we already use `fail` in the script, so it's better to use that version.

arunk-s:
Oh, Okay. One addition: Since the `fail` is going to be executed if `grep` expression matches(exits successfully), it won't give any indication to user that the go version should be updated.
So maybe we do this:
```bash
echo $go_version | grep -E -q --regexp="go1\.[0-4]" \
&& echo -e "installed version of go is ${go_version}\nwe need at least version 1.5" && fail
```

jvehent:
that sounds good

arunk-s:
Pull request updated for the final version.

jvehent:
Excellent! thanks for the patch.

PR #191, opened 2016-03-08
[minor] Fixes #115

### LINKED ISSUE #115: [mig client] Add option for target agents from results of previous action (Bugzilla #1141910)
Migrated from https://bugzilla.mozilla.org/show_bug.cgi?id=1141910

On 2015-03-10 20:56:03 -0700, Julien Vehent [:ulfr] wrote:

> Add options that automatically implements http://mig.mozilla.org/doc/cheatsheet.rst.html#target-agents-that-found-results-in-a-previous-action
> 
> mig file -t "id IN ( \
>     SELECT agentid FROM commands, json_array_elements(commands.results) AS r \
>     WHERE commands.actionid = 12345 AND r#>>'{foundanything}' = 'true')" \
> -path /etc/passwd -content "^spongebob"
> 
> For example:
> * --target-foundanything <actionid> would limit the targets to agents that returned positive results in <actionid>
> * --target-foundnothing <actionid> would do the opposite and target agents that did not return positive results in <actionid>
> 
> This option should concatenate with the regular -t option such that an investigator could set both and filter the results even further.

On 2015-03-10 20:56:41 -0700, Julien Vehent [:ulfr] wrote:

> Implementation should take the form of two new options in the command line client at https://github.com/mozilla/mig/blob/master/src/mig/client/cmd/main.go#L75

arunk-s:
Add `-target-foundanything` and `-targetfoundnothing` flags to mig client

### DIFF
--- a/client/mig/main.go
+++ b/client/mig/main.go
@@ -45,6 +45,9 @@ usage: %s <module> <global options> <module parameters>
 		* proxied linux agents:  -t "queueloc LIKE 'linux.%%' AND environment->>'isproxied' = 'true'"
 		* agents operated by IT: -t "tags#>>'{operator}'='IT'"
 		* run on local system:	 -t local
+-target-found <action ID> targets agents that have found results in a previous action. ex: -target-found 123456
+-target-foundnothing <action ID> targets agents that did not find results in a previous action.
+		Both -target-found and -target-foundnothing cannot be used simultaneously.
 -v		verbose output, includes debug information and raw queries
 -V		print version
 -z <bool>       compress action before sending it to agents
@@ -69,17 +72,18 @@ func continueOnFlagError() {
 
 func main() {
 	var (
-		conf                                           client.Configuration
-		cli                                            client.Client
-		err                                            error
-		op                                             mig.Operation
-		a                                              mig.Action
-		migrc, show, render, target, expiration, afile string
-		printAndExit                                   bool
-		verbose, showversion                           bool
-		compressAction                                 bool
-		modargs                                        []string
-		run                                            interface{}
+		conf                                    client.Configuration
+		cli                                     client.Client
+		err                                     error
+		op                                      mig.Operation
+		a                                       mig.Action
+		migrc, show, render, target, expiration string
+		afile, targetfound, targetnotfound      string
+		printAndExit                            bool
+		verbose, showversion                    bool
+		compressAction                          bool
+		modargs                                 []string
+		run                                     interface{}
 	)
 	defer func() {
 		if e := recover(); e != nil {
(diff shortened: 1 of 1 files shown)


client/mig/main.go
```diff
@@ -223,6 +230,20 @@ func main() {
 	for _, arg := range os.Args[1:] {
 		a.Name += arg + " "
 	}
+
+	if targetfound != "" && targetnotfound != "" {
+		panic("Both -target-foundanything and -target-foundnothing cannot be used simultaneously")
+	}
+	if targetfound != "" {
+		targetQuery := fmt.Sprintf(`id IN (select agentid from commands, json_array_elements(commands.results) as 
+			r where actionid=%s and r#>>'{foundanything}' = 'true')`, targetfound)
```
jvehent: we probably don't want a newline in here, as the string will be added to the json

client/mig/main.go
```diff
@@ -223,6 +230,20 @@ func main() {
...
+		target = targetQuery + " AND " + target
+	}
+	if targetnotfound != "" {
+		targetQuery := fmt.Sprintf(`id NOT IN (select agentid from commands, json_array_elements(commands.results) as 
+			r where actionid=%s and r#>>'{foundanything}' = 'true')`, targetnotfound)
```
jvehent: same comment about newline

client/mig/main.go
```diff
@@ -94,6 +99,8 @@ func main() {
 	fs.StringVar(&show, "show", "found", "type of results to show")
 	fs.StringVar(&render, "render", "text", "results rendering mode")
 	fs.StringVar(&target, "t", fmt.Sprintf("status='%s' AND mode='daemon'", mig.AgtStatusOnline), "action target")
+	fs.StringVar(&targetfound, "target-foundanything", "", "Limit targets to agents that returned positive results")
+	fs.StringVar(&targetnotfound, "target-foundnothing", "", "Limit targets to agents that did not returned positive results")
```
jvehent: These explanations are vague. What about `targets agents that have found results in a previous action. ex: -target-found 123456`.

client/mig/main.go
```diff
@@ -94,6 +99,8 @@ func main() {
 	fs.StringVar(&show, "show", "found", "type of results to show")
 	fs.StringVar(&render, "render", "text", "results rendering mode")
 	fs.StringVar(&target, "t", fmt.Sprintf("status='%s' AND mode='daemon'", mig.AgtStatusOnline), "action target")
+	fs.StringVar(&targetfound, "target-foundanything", "", "Limit targets to agents that returned positive results")
```
jvehent: we can remove the `anything` part of the string, `target-found` is enough.

client/mig/main.go
arunk-s: Oh, I thought that by line wrapping you meant that I should divide the line into two.
Should I rather be doing this:
```go
targetQuery := fmt.Sprintf(`id IN (select agentid from commands, json_array_elements(commands.results) as ` +
`r where actionid=%s and r#>>'{foundanything}' = 'true')`, targetfound) 
```
which divide the line into two and also doesn't add a newline.

client/mig/main.go
arunk-s: Yeah. That makes more sense. I'll update the code to reflect that.

jvehent:
r+, thanks for the patch :)

jvehent:
I'm fixing bugs with this code in another commit.

PR #192, opened 2016-03-14
[minor] Fixes #144: Add NeighborIP functionality on linux

### LINKED ISSUE #144: netstat NeighborIP is not used
The NeighborIP parameter in the netstat module does not appear to be used; the command line provides the ability to search using -ni but it doesn't look like this incoming parameter is processed anywhere currently.

arunk-s:


modules/netstat/netstat_linux.go
```diff
@@ -584,3 +584,121 @@ func procNetNSEntries(fname string, nsEntrych chan string, nsErrch chan error) {
...
+		ipnet = new(net.IPNet)
+		ipnet.IP = ip
+		ipnet.Mask = net.CIDRMask(net.IPv4len*8, net.IPv4len*8)
+	}
+	for _, arpent := range lns {
+		fields := strings.Fields(arpent.line)
+		if len(fields) < 4 {
+			continue
+		}
+        remoteIP := net.ParseIP(fields[0])
```
jvehent: indent

jvehent:
It looks good! and local test passes:
```
$ mig netstat -t local -ni 172.21.0.1
found neighbor IP 172.21.0.1 64:66:b3:02:76:e6 for netstat neighborIP:'172.21.0.1'
```
Do you think you could add the macos and windows support as well, or should that be a different patch?

arunk-s:
@jvehent currently I don't have access to mac (as I don't own one) and my laptop have currently ubuntu only. So I think that should be added in a later patch :)

But if you want me to work on them, I can try to setup a VM (btw I've heard that macos in VM are tricky ;) ).

arunk-s:
@jvehent Also PR is updated with new changes.

jvehent:
Patch looks good, thanks a lot!

PR #206, opened 2016-04-05
resolve deadlock in scheduler periodic run

ameihm0912:
Resolves a deadlock condition in scheduler periodic run that prevents any future periodic runs from occurring + minor fix related to periodic error checking


jvehent:
r+
:cry: bugs

PR #207, opened 2016-04-05
various fixes related to loader testing

ameihm0912:


logging_posix.go
```diff
@@ -95,6 +98,83 @@ func (l Log) Debug() (mlog Log) {
...
+	}
+	if fi.Size() < r.maxBytes {
+		return
+	}
+	// Rotate the log and reinitialize it
+	// If the old file already exists remove it
+	rotatefile := r.filename + ".1"
+	_, err = os.Stat(rotatefile)
+	if err == nil || !os.IsNotExist(err) {
+		err = os.Remove(rotatefile)
```
jvehent: Another option would be to suffix the file with `date +%Y%m%d%H%M` and keep many versions, but I'm fine with the current approach too as it prevents filing up disks.

mig-loader/loader.go
```diff
@@ -223,7 +223,8 @@ func fetchAndReplace(entry mig.BundleDictionaryEntry, sig string) (err error) {
 	//
 	// Append .loader to the file name to use as the staged file path.
 	reppath := entry.Path + ".loader"
-	fd, err := os.OpenFile(reppath, os.O_CREATE|os.O_TRUNC|os.O_WRONLY, 0700)
+	fd, err := os.OpenFile(reppath, os.O_CREATE|os.O_TRUNC|os.O_WRONLY,
+		entry.Perm)
```
jvehent: nit: looks better without the newline

mig-loader/loader.go
```diff
@@ -346,6 +347,29 @@ func checkEntry(entry mig.BundleDictionaryEntry) (err error) {
...
+		if e := recover(); e != nil {
+			err = fmt.Errorf("entryPermCheck() -> %v", e)
+		}
+	}()
+	var fi os.FileInfo
+	fi, err = os.Stat(entry.Path)
+	if err != nil {
+		panic(err)
+	}
+	if fi.Mode() != entry.Perm {
```
jvehent: is it faster to skip the `Stat` and `Mode` call and just do the `Chmod` regardless?

conf/mig-agent-conf.go.inc
```diff
@@ -34,6 +34,10 @@ var DISCOVERAWSMETA = true
 // and exits. this mode is used to run the agent as a cron job, not a daemon.
 var CHECKIN = false
 
+// how often the agent will refresh its environment. if 0 agent
+// will only update environment at initialization.
+var REFRESHENV time.Duration = 0
```
jvehent: what about 1800 as a default?

mig-agent/context.go
```diff
@@ -285,6 +286,19 @@ func initAgentID(orig_ctx Context) (ctx Context, err error) {
 			panic(err)
 		}
 	}
+	// Make sure the obtained queue location matches the format that we expect, if
+	// it doesn't create a new one
+	mtch, err := regexp.Match("^\\S+\\.\\S+$", id)
```
jvehent: This is going to break backward compat and all agents will create a new queueloc.
I _think_ this won't have too much impact, but it's friday and I'm drinking a 10% beer...

mig-agent/context.go
```diff
@@ -304,6 +318,11 @@ func createIDFile(ctx Context) (id []byte, err error) {
 	sid += strconv.FormatUint(uint64(r.Int63()), 36)
 	sid += strconv.FormatUint(uint64(r.Int63()), 36)
 	sid += strconv.FormatUint(uint64(r.Int63()), 36)
+	sid += "."
```
jvehent: I dot (see what I did here) this is needed.

database/agents.go
```diff
@@ -117,8 +117,14 @@ func (db *DB) InsertAgent(agt mig.Agent) (err error) {
...
-	_, err = db.c.Exec(`UPDATE agents SET status=$1, heartbeattime=$2 WHERE id=$3`,
-		mig.AgtStatusOnline, agt.HeartBeatTS, agt.ID)
+	jEnv, err := json.Marshal(agt.Env)
+	if err != nil {
+		err = fmt.Errorf("Failed to marshal agent environment: '%v'", err)
+		return
+	}
+	_, err = db.c.Exec(`UPDATE agents SET status=$1, heartbeattime=$2,
+		environment=$3, name=$4 WHERE id=$5`,
+		mig.AgtStatusOnline, agt.HeartBeatTS, jEnv, agt.Name, agt.ID)
```
jvehent: r-
This is going to overwrite existing environments. We need to keep history (useful for forensics).
It would be useful for the agent to send a flag alongside the heartbeat when its environment has changed, then the scheduler would know to store a new row instead of updating the existing one.

jvehent:
Done. Several comments, but nothing critical except for the env update in DB.

database/agents.go
ameihm0912: @jvehent Yeah. We could mark the agent immediately as offline on update, or potentially introduce a new status (e.g., "superseded") for this. Since we search by queue and PID and != offline don't think we can just introduce a new row here alone. Going immediately to offline I don't really like as it bypasses the normal agent lifecycle, but at that point maybe its the correct status for the agent.

Another option is adding a new relation to the agents table, to track env/tag/name over time and removing those columns from agents, this seems more complicated than is required though.

PR #208, opened 2016-04-14
fixes to scheduler

ameihm0912:
Removes upgrade module code in the scheduler.

Ensure we drain DetectDupAgents even if the option to kill duplicate agents is disabled in the scheduler configuration. There are a few places in the code we write to this channel, currently if the option is disabled the receiving end will not be spawned. This leads to routines in the scheduler blocking when they write here. We already check the configuration in the agent kill function in the scheduler, so just drain the channel and call this function and it is handled there depending on the setting.


jvehent:
I always hated that update code. Thanks for getting rid of it!

r+

PR #214, opened 2016-04-22
[minor/bug] fix to build windows agent

yvesago:
Hello,

here is my first pull request for mig ;-)

It contains some simple fixes to allow building a windows client with the new agentcontext feature and fix the old bug in logging_windows.
Building module memory still fail.

### DIFF
--- a/logging_windows.go
+++ b/logging_windows.go
@@ -13,7 +13,7 @@ import (
 	"os"
 	"sync"
 
-	"code.google.com/p/winsvc/eventlog"
+	"golang.org/x/sys/windows/svc/eventlog"
 )
 
 const (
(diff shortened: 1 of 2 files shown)


logging_windows.go
```diff
@@ -13,7 +13,7 @@ import (
 	"os"
 	"sync"
 
-	"code.google.com/p/winsvc/eventlog"
+    "golang.org/x/sys/windows/svc/eventlog"
```
ameihm0912: it looks like we have some indenting issues here -- could you run gofmt on your files changed, just to make sure we stay consistent with indenting and so on?

ameihm0912:
@yvesago thanks for grabbing this one. It looks good to me, if we could just fix up any formatting issues via gofmt.

yvesago:
This one should be better (if fixup works as I hope)

ameihm0912:
Looks good, thanks!

PR #219, opened 2016-04-28
Fixes in netstat module

ameihm0912:
Corrects an issue with Darwin -ci flag searches, and a bug in Linux IPv6 -ci searches. Also adds support for scanning IPv4 mapped IPv6 addresses.


jvehent:
All looks good. r+

PR #220, opened 2016-04-28
[minor/bug] pathwalh error on symlink directory trailing  slash to fix #213

### LINKED ISSUE #213: Cannot use C:\\ as search path
When investigating a Windows machine, using C:\\ as the search path does not work. Using multiple C:\\<folder-name> does.

yvesago:
A light patch to fix #213 
It doesn't add ending separator on symlink and doesn't add symlinked directories to subdirs

### DIFF
--- a/modules/file/file.go
+++ b/modules/file/file.go
@@ -833,18 +833,22 @@ func (r *run) pathWalk(path string, roots []string) (traversed []string, err err
 		// loop over the content of the directory
 		for _, dirEntry := range dirContent {
 			entryAbsPath := path
-			// append path separator if missing
-			if entryAbsPath[len(entryAbsPath)-1] != os.PathSeparator {
+			// append path separator if missing & not a symlink
+			// (maybe useless, directories are tested in next block)
+			if entryAbsPath[len(entryAbsPath)-1] != os.PathSeparator && (dirEntry.Mode()&os.ModeSymlink != os.ModeSymlink) {
 				entryAbsPath += string(os.PathSeparator)
 			}
 			entryAbsPath += dirEntry.Name()
 			// this entry is a subdirectory, keep it for later
 			if dirEntry.IsDir() {
-				// append trailing slash
-				if entryAbsPath[len(entryAbsPath)-1] != os.PathSeparator {
-					entryAbsPath += string(os.PathSeparator)
+				// if not symlinked directory, don't put symlinked directory in subdirectories to follow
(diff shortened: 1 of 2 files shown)


modules/file/file.go
```diff
@@ -835,16 +835,22 @@ func (r *run) pathWalk(path string, roots []string) (traversed []string, err err
 			entryAbsPath := path
 			// append path separator if missing
 			if entryAbsPath[len(entryAbsPath)-1] != os.PathSeparator {
-				entryAbsPath += string(os.PathSeparator)
+				// if not a symlink
+				if dirEntry.Mode()&os.ModeSymlink != os.ModeSymlink {
+					entryAbsPath += string(os.PathSeparator)
+				}
```
jvehent: Could you rewrite this as a single if condition?
```go
if (entryAbsPath[len(entryAbsPath)-1] != os.PathSeparator) && (dirEntry.Mode()&os.ModeSymlink != os.ModeSymlink) {
    entryAbsPath += string(os.PathSeparator)
}
```

modules/file/file.go
```diff
@@ -835,16 +835,22 @@ func (r *run) pathWalk(path string, roots []string) (traversed []string, err err
...
 			}
 			entryAbsPath += dirEntry.Name()
 			// this entry is a subdirectory, keep it for later
 			if dirEntry.IsDir() {
-				// append trailing slash
-				if entryAbsPath[len(entryAbsPath)-1] != os.PathSeparator {
-					entryAbsPath += string(os.PathSeparator)
+				// don't follow directory symlink
+				if dirEntry.Mode()&os.ModeSymlink != os.ModeSymlink {
```
jvehent: Would this ever match given the if condition with requires `dirEntry.IsDir()` ?

jvehent:
I tested this on windows 7 and it works, so :+1: to that!
In addition to addressing the couple comments I put on the commit, could you:
* remove the `merge` commit through a rebase
* reword the `minor/bug` commit to `medium/bug` and indicate in the commit message that it solves a Windows issue

modules/file/file.go
yvesago: Sure but I have a doubt in the usefulness of this bloc.

modules/file/file.go
yvesago: Yes, this allow to not put symlinked sub directories in subdir and fix real dir path. But I think that the other bloc in 838 is useless.

PR #223, opened 2016-05-02
[medium/bug] windows bug in pathwalk, fixes #213

### LINKED ISSUE #213: Cannot use C:\\ as search path

yvesago:
Here the right Monday patch  ;-)
Block in line 838 is useful for first path other subdirs path are fixed in 844


jvehent:
Looks good to me. Thanks for the patch! :beer:

PR #224, opened 2016-05-02
[medium/bug] fix windows agent environment

yvesago:
use wmic instead of the long and translated systeminfo


jvehent:
Testing on my local machine and works as expected. Thanks for the patch :+1:

Issue #197, opened 2016-03-24
Debian package for mig-loader gives maintainer-address-missing error on installation


gene1wood:
The [debian package of the mig-loader](https://s3.amazonaws.com/mozopsecrepo2/deb/Generic/mig-loader_20160304-0.ae47410.stage_amd64.deb) gives the following error on installation.

```
Lintian check results for /path/to/mig-loader_20160304-0.ae47410.stage_amd64.deb:
E: mig-loader: maintainer-address-missing Mozilla OpSec
```

![untitled window_120](https://cloud.githubusercontent.com/assets/1134034/14031691/328b4dd2-f1cb-11e5-87a8-55be17355762.png)

This error is given if the mig-loader key is present in `/etc/mig/mig-loader.key` or not.

PR #231, opened 2016-05-16
Some loader documentation updates

ameihm0912:
First updates to loader documentation...

More to come...

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -178,6 +178,18 @@ deb-loader: mig-loader
 		--architecture $(FPMARCH) -v $(BUILDREV) \
 		-s dir -t deb .
 
+rpm-loader: mig-loader
+	rm -fr tmp
+	$(INSTALL) -s -D -m 0755 $(BINDIR)/mig-loader tmp/sbin/mig-loader
+	$(INSTALL) -D -m 0644 LICENSE tmp/usr/share/doc/mig-loader/copyright
+	$(MKDIR) -p tmp/var/lib/mig
+	$(MKDIR) -p tmp/etc/mig
+	fpm -C tmp -n mig-loader --license GPL --vendor mozilla \
+		--description "Mozilla InvestiGator Agent Loader\nAgent loader binary" \
+		-m "Mozilla <noreply@mozilla.com>" --url http://mig.mozilla.org \
+		--architecture $(FPMARCH) -v $(BUILDREV) \
+		-s dir -t rpm .
+
 dmg-agent: mig-agent
 ifneq ($(OS),darwin)
 	echo 'you must be on MacOS and set OS=darwin on the make command line to build an OSX package'
--- a/doc/agent.rst
+++ b/doc/agent.rst
@@ -61,6 +61,14 @@ Below is a sample heartbeat message from a linux agent stored in
 		"version": "201407310027+bcbdd94.prod"
 	}
 
+It's possible on an endpoint information in the environment will change. For
+example, a new IP address could be assigned via DHCP. The agent periodically
+checks the system; if changes to the environment are detected the heartbeat
+message will automatically be updated to include those changes. The frequency
+environment checks occur can be controlled through the ``refreshenv``
+configuration option in the agent configuration file, or the ``REFRESHENV``
+variable in the agent built-in configuration.
+
 Check-In mode
 ~~~~~~~~~~~~~
 In infrastructure where running the agent as a permanent process is not
--- a/doc/agent.rst
+++ b/doc/agent.rst
@@ -63,13 +63,16 @@ Below is a sample heartbeat message from a linux agent stored in
 		"version": "201407310027+bcbdd94.prod"
 	}
 
-It's possible on an endpoint information in the environment will change. For
-example, a new IP address could be assigned via DHCP. The agent periodically
-checks the system; if changes to the environment are detected the heartbeat
-message will automatically be updated to include those changes. The frequency
-environment checks occur can be controlled through the ``refreshenv``
-configuration option in the agent configuration file, or the ``REFRESHENV``
-variable in the agent built-in configuration.
+The agent sends information about the OS configuration and it's environment
+to the scheduler periodically. This includes information like the hostname
+of the system it is running on, IP addresses assigned, AWS instance related
+information, and others. It's possible on an endpoint this changes while the
+agent is running. For example, a new IP address could be assigned via DHCP.
+The agent periodically checks the system; if changes to the environment
+are detected the heartbeat message will automatically be updated to include
+those changes. The frequency environment checks occur can be controlled
+through the ``refreshenv`` configuration option in the agent configuration
+file, or the ``REFRESHENV`` variable in the agent built-in configuration.
 
 Check-In mode
 ~~~~~~~~~~~~~
--- a/doc/agent.rst
+++ b/doc/agent.rst
@@ -10,7 +10,9 @@ Initialization process
 The agent tries to be as autonomous as possible. One of the goal is to ship
 agents without requiring external provisioning tools, such as Chef or Puppet.
 Therefore, the agent attempts to install itself as a service, and also supports
-a builtin upgrade protocol (described in the next section).
+optional automatic upgrades via the `mig-loader`_ companion program.
+
+.. _`mig-loader`: loader.rst
 
 As a portable binary, the agent needs to detect the type of operating system
 and init method that is used by an endpoint. Depending on the endpoint,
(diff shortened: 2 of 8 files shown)


ameihm0912:
@jvehent want to have a look at this when you have time and see if it looks OK to merge?

Note: API documentation is coming as well, not in here yet

jvehent:
Mind if I wait for the API doc? I'd rather do one pass.

ameihm0912:
No problem should be ready soon

ameihm0912:
@jvehent any thoughts on this one?

jvehent:
@gdestuynder : r?

Makefile
```diff
@@ -178,6 +178,18 @@ deb-loader: mig-loader
 		--architecture $(FPMARCH) -v $(BUILDREV) \
 		-s dir -t deb .
 
+rpm-loader: mig-loader
+	rm -fr tmp
+	$(INSTALL) -s -D -m 0755 $(BINDIR)/mig-loader tmp/sbin/mig-loader
+	$(INSTALL) -D -m 0644 LICENSE tmp/usr/share/doc/mig-loader/copyright
```
gdestuynder: just a nitpick, why not keep LICENSE? (since its rather standard). r+ either way.

doc/agent.rst
```diff
@@ -10,7 +10,9 @@ Initialization process
 The agent tries to be as autonomous as possible. One of the goal is to ship
 agents without requiring external provisioning tools, such as Chef or Puppet.
 Therefore, the agent attempts to install itself as a service, and also supports
-a builtin upgrade protocol (described in the next section).
```
gdestuynder: that means no more builtin upgrade module right?

doc/agent.rst
```diff
@@ -61,6 +63,14 @@ Below is a sample heartbeat message from a linux agent stored in
 		"version": "201407310027+bcbdd94.prod"
 	}
 
+It's possible on an endpoint information in the environment will change. For
```
gdestuynder: It's possible that information on an endpoint will change depending on the environment ? (please rephrase)

doc/agent.rst
```diff
@@ -249,104 +259,55 @@ and copies them into ``Message.Parameters``. It then sets ``Message.Class`` to
...
+        | matches   |        |                     |
+        | manifest? +------->|                     |
+        |                    |                     |
+        |  8. install agent  |                     |
+        |<-------------------|                     |
+        |                    |                     |
+        |  9. stop old agent |                     |
+        |<-------------------|                     |
+        |                    |                     |
+        | 10. start new      |                     |
```
gdestuynder: what happens if the new agent fails to start?

Makefile
ameihm0912: This change was a lintian thing (https://lintian.debian.org/). Some package mgr frontends seem to run this and return errors if the lintian passes don't check.

This file is required in the checks so we just use that filename in this case (see #197).

doc/agent.rst
ameihm0912: Yeah; I believe PR catches all the references to the upgrade module in docs, but we probably still have some code around it.

doc/agent.rst
ameihm0912: Will fix this one up

doc/agent.rst
ameihm0912: Start may be the wrong term in the doc here; what actually happens is the loader starts the agent, and the agent installs itself as a service to run and exits. In normal circumstances, after this the service mgr will start the agent up (systemd, upstart, etc).

If for some reason this fails, the intent is since the loader is in cron/systemd timer it would run again. When that happens it will detect the agent isn't running and repeat the process.

This would correct any intermittent sort of error, if it's something going wrong we can't recover it unless we deploy new loader/agent code. This happens before the start, so if the error is resolved we can fix it and it would correct on next run after the manifest is updated.

doc/agent.rst
jvehent: The environment of an endpoint may change over time.

doc/agent.rst
```diff
@@ -249,104 +259,55 @@ and copies them into ``Message.Parameters``. It then sets ``Message.Class`` to
...
+program mig-loader. Using mig-loader is optional; you don't need to use
+mig-loader in your environment if you want to upgrade agents yourself.
+
+The following is a high level diagram of how the loader interacts with the
+API and the agent during the upgrade process.
+
+::
+
+        Agent                Loader              API
+        +---+                +----+             +--+
```
jvehent: Maybe add a header to this showing both Agent and Loader are located on the endpoint, otherwise people may wonder where the loader is placed.

```
        /--------Endpoint---------\
        Agent                Loader              API
        +---+                +----+             +--+
```

doc/agent.rst
```diff
@@ -249,104 +259,55 @@ and copies them into ``Message.Parameters``. It then sets ``Message.Class`` to
...
+        | 3. valid  +--------|                     |      | record in database
+        | manifest  |        |                     |<-----+
+        | sig?      +------->|                     |
+        |                    |                     |
+        | 4. does   +--------|                     |
+        | current   |        |                     |
+        | agent     |        |                     |
+        | match?    +------->|                     |
+        |                    |                     |
+        |                    | 5. fetch new agent  |
```
jvehent: fetch files referenced in manifest?

doc/api.rst
```diff
@@ -862,6 +862,289 @@ Find the last 10 commands signed by an investigator identified by name.
...
+                        "data": [
+                            {
+                                "name": "loader",
+                                "value": {
+                                    "AgentName": "corbomite.internal",
+                                    "Enabled": true,
+                                    "ID": 12,
+                                    "Key": "",
+                                    "LastSeen": "2016-05-17T14:10:03.041024-05:00",
+                                    "Name": "corbomite.internal"
```
jvehent: Are those keys really capitalized? If so, can they be lowercased to be consistent with other API responses?

doc/api.rst
```diff
@@ -862,6 +862,289 @@ Find the last 10 commands signed by an investigator identified by name.
...
+        - `loaderkey`: New key for loader instance
+* Authentication: X-PGPAUTHORIZATION
+* Response Code: 200 OK
+* Reponse: Collection+JSON
+
+POST /api/v1/loader/new/
+~~~~~~~~~~~~~~~~~~~~~~~~
+* Description: Create a new loader instance
+* Parameters: (POST body)
+	- `loader`: JSON marshaled mig.LoaderEntry data
```
jvehent: marshalled

doc/api.rst
```diff
@@ -862,6 +862,289 @@ Find the last 10 commands signed by an investigator identified by name.
...
+                    }
+                ],
+                "template": {},
+                "version": "1.0"
+            }
+        }
+
+POST /api/v1/manifest/agent/
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+* Description: Returns a manifest for consumption by mig-loader on an agent endpoint
```
jvehent: `Returns` for a POST request?

doc/api.rst
```diff
@@ -862,6 +862,289 @@ Find the last 10 commands signed by an investigator identified by name.
...
+            "collection": {
+                "error": {},
+                "href": "http://api.mig.example.net:1664/api/v1/manifest/fetch/",
+                "items": [
+                    {
+                        "data": [
+                            {
+                                "name": "content",
+                                "value": {
+                                    "data": "<file content...>",
```
jvehent: Could we specify the encoding of the file content? I think it's base64, it is compressed?

jvehent: Actually, might be nice to have the encoding defined in `value.encoding` in the doc directly.

doc/configuration.rst
```diff
@@ -728,6 +728,31 @@ created with ID 2". We can view the details of this new investigator by entering
 	created  2015-09-09 09:53:28.989481 -0400 EDT
 	modified 2015-09-09 09:53:28.989481 -0400 EDT
 
+Recent versions of MIG have introduced the concept of investigators that can be
```
jvehent: I wouldn't bother with "recent".
`MIG supports two levels of access its users: investigators and administrators.`

By the way, should administrators be allowed to launch investigations, or are the two roles mutually exclusive?

doc/configuration.rst
```diff
@@ -728,6 +728,31 @@ created with ID 2". We can view the details of this new investigator by entering
...
+::
+
+    $ sudo -u postgres psql mig
+    psql (9.4.7)
+    Type "help" for help.
+    
+    mig=# UPDATE investigators SET isadmin = true WHERE id=2;
+    UPDATE 1
+    mig=# \q
+    $
```
jvehent: I don't think we should ask people to touch the DB directly. Any other way to do this? I know it's a chicken and egg problem for the first admin...

jvehent: This phase of the configuration happens before the API authentication is enabled. Maybe we could allows any investigator to set or unset the administrator bit when the api auth is disabled, which would solved the chicken and egg issue.
Then we could just manage administrators in the API/Console.
Thoughts?

doc/configuration.rst
```diff
@@ -754,6 +779,30 @@ and thus a compromise of the API doesn't leak credentials of investigators.
...
+
+With mig-loader, instead of installing the agent on the systems you want to run
+the agent on, you would install only mig-loader. mig-loader is a small binary
+intended to be run from a periodic system such as cron. mig-loader will then
+look after fetching the agent and installing it if it does not exist on the system,
+and will look after upgrading the agent automatically if you want to publish new
+agent updates. The upgrades can be controlled by a MIG administrator through the
+MIG API and console tools.
+
+**Note:** mig-loader is a new feature. It's use is optional.
```
jvehent: I don't think the note is necessary.

doc/configuration.rst
```diff
@@ -754,6 +779,30 @@ and thus a compromise of the API doesn't leak credentials of investigators.
...
+
+For information on the loader, see `MIG LOADER`_ documentation. If you wish to
```
jvehent: Remove first sentence, redundant.

doc/loader.rst
```diff
@@ -5,128 +5,539 @@ MIG LOADER
...
+new agent code is available, making sure it's signed and safe to install, and
+handling automatic agent/configuration upgrades.
+
+In environments with a mature configuration management setup (such as via
+puppet or ansible) MIG loader may not be required. The configuration management
+system can periodically push updates out to required systems.
+
+In other environments, the loader can help keep things up to date. As an example
+this could be in cases where you want to keep MIG up to date on workstations that
+are not centrally managed. Even in cases with a configuration management system in
```
jvehent: "keep things up to date, for example on workstations and laptops that are not centrally managed."

doc/loader.rst
```diff
@@ -5,128 +5,539 @@ MIG LOADER
...
+place, you may want to use MIG loader to provide more fine-grained control over
+updates to agents.
+
+Here we describe how the loader works, and provide an example on how it is setup
+and used to provision and continuously upgrade a MIG environment.
```
jvehent: "upgrade MIG Agents."
The rest of the env isn't upgraded by the loader, although that'd be neat :)

doc/loader.rst
```diff
@@ -5,128 +5,539 @@ MIG LOADER
...
+systems. If you wanted to upgrade the agent, you would need to upgrade the agent
+package installed on the systems.
+
+With mig-loader, instead of installing the agent on the systems you want to run
+the agent on, you would install only mig-loader. mig-loader is a small binary
+intended to be run from a periodic system such as cron. mig-loader will then
+look after fetching the agent and installing it if it does not exist on the system,
+and will look after upgrading the agent automatically if you want to publish new
+agent updates. The upgrades can be controlled by a MIG administrator through the
+MIG API and console tools.
```
jvehent: That's a bit redundant with the previous section, maybe merge the two?

doc/loader.rst
```diff
@@ -5,128 +5,539 @@ MIG LOADER
...
+
+Key components
+--------------
+Several components within MIG are responsible for supporting loader related
+deployment. Some are dedicated to this, others have additional functionality
+that is used.
+
+mig-loader
+~~~~~~~~~~
+mig-loader is the component responsible for upgrading and managing the agent and
```
jvehent: "mig-loader runs on an endpoint to manage its version of mig-agent".

doc/loader.rst
```diff
@@ -5,128 +5,539 @@ MIG LOADER
...
+associated agent configuration files on a system. It would run on a host you would
+like the agent deployed on. mig-loader looks after periodically checking if new
+agents are available by contacting the API, validating signatures on new agent
+manifests, and installing new agent code and configuration data when available. It
+also looks after management related actions such as restarting the agent if new
+code has been installed.
```
jvehent: "if a new version has been installed"

doc/loader.rst
```diff
@@ -5,128 +5,539 @@ MIG LOADER
...
+
+mig-api
+~~~~~~~
+In the general sense, the MIG API is used by clients to send actions to agents and
```
jvehent: "the API is the control interface of MIG used by investigators to interact with the platform."

doc/loader.rst
```diff
@@ -5,128 +5,539 @@ MIG LOADER
...
 
-* Configure the loader and build it
-* Create a manifest containing the agent and other support files, and make it avialable through the API
-* Install the loader on systems you wish to run the agent on
+About manifest signatures
+-------------------------
+When the loader asks the API for the current version of the agent that should be
+installed, the API will respond with a signed manifest. The manifest is signed by
+MIG administrators when it is uploaded to the API (discussed later) using the
+administrators GPG key. The loader is built with the GPG public key of the
```
jvehent: `s/GPG/PGP/g`
We actually don't use GPG, we use Golang OpenPGP ;)

doc/loader.rst
```diff
@@ -5,128 +5,539 @@ MIG LOADER
...
+::
+
+    $ make mig-loader LOADERCONF=conf/mig-loader-myenv.go.inc
+    mkdir -p bin/linux/amd64
+    if [ ! -r conf/mig-loader-myenv.go.inc ]; then echo "conf/mig-loader-myenv.go.inc configuration file does not exist" ; exit 1; fi
+    # test if the loader configuration variable contains something different than the default value
+    # and if so, replace the link to the default configuration with the provided configuration
+    if [ conf/mig-loader-myenv.go.inc != "conf/mig-loader-conf.go.inc" ]; then rm mig-loader/configuration.go; cp conf/mig-loader-myenv.go.inc mig-loader/configuration.go; fi
+    GOOS=linux GOARCH=amd64 GO15VENDOREXPERIMENT=1 go build  -o bin/linux/amd64/mig-loader -ldflags "-X mig.ninja/mig.Version=20160512-0.9fe5f23.dev" mig.ninja/mig/mig-loader
+    $
```
jvehent: nit: `$`

doc/loader.rst
```diff
@@ -5,128 +5,539 @@ MIG LOADER
...
+::
+
+    mig> create manifest
+    Entering manifest creation mode.
+    Please provide the name of the new manifest
+    name> a new manifest
+    Name: 'a new manifest'
+    Please provide loader targeting string for manifest.
+    target> env#>>'{os}'='linux'
+    Target: 'env#>>'{os}'='linux''
```
jvehent: All environments have the `os` flag set so you could use `env->>'os'='linux'` instead.

doc/loader.rst
```diff
@@ -5,128 +5,539 @@ MIG LOADER
...
+The following example shows how you would add a user and set the required
+permissions on your relay, using ``rabbitmqctl``.
+
+::
+
+    # rabbitmqctl add_user corbomite.internal vw1NQs9F3wuZx1kSUhZaQzZ0
+    Creating user "corbomite.internal" ...
+    # rabbitmqctl set_permissions -p mig corbomite.internal '^mig\.agt\..*$' '^(toschedulers|mig\.agt\..*)$' '^(toagents|mig\.agt\..*)$'
+    Setting permissions for user "corbomite.internal" in vhost "mig" ...
+    #
```
jvehent: We still need to fix this. MIG Administrators won't have access to `rabbitmqctl`.

doc/loader.rst
```diff
@@ -5,128 +5,539 @@ MIG LOADER
...
+    comparing configuration /etc/mig/mig-agent.cfg
+    we have not found
+    they have d51a2e9d955aaca94e88159ad6235cbaccf9680f0d8e82dcee0f2f0f0df83038
+    refreshing configuration
+    fetching file from https://my.mig.api.url:1664/api/v1/manifest/fetch/
+    validating staged file signature
+    renaming existing file
+    installing staged file
+    comparing agentcert /etc/mig/agent.crt
+    we have not found
```
jvehent: `we have not found` ....
proposal:
* your underwear
* the cat
* any beer in the fridge
* the answer to life

jvehent:
Looks like I was too late for the review :cry: 
Still I put in my comments. Overall it's a good doc. It pointed out a couple areas where I think we need to improve usability, but those can be tackled as future improvements.

Would appreciate if you handle the comments :)

doc/api.rst
ameihm0912: Good catch this is a bug, we are missing JSON tags here: https://github.com/mozilla/mig/blob/master/loader.go#L16-L23

PR #232, opened 2016-05-22
[minor] support filtering by version in pkg module

### LINKED ISSUE #182: add -version flag to pkg module
It would be nice to have a -version flag with a dependency on the -name flag added.
This would allow investigator to query for specific packages across environments.

For instance, I'm looking for a vulnerable package, and only want to see the hosts with that vulnerable package.

In MIG's current iteration it returns all package versions that share the same -name.

ameihm0912:
Closes #182

### DIFF
--- a/modules/pkg/paramscreator.go
+++ b/modules/pkg/paramscreator.go
@@ -24,7 +24,11 @@ func printHelp(isCmd bool) {
 %sname <regexp>     - OS package search
                     ex: name linux-image
 		    query for installed OS packages matching expression
-`, dash)
+
+%sversion <regexp>  - Version string search
+                    ex: version ^1\..*
+		    optionally filter returned packages to include version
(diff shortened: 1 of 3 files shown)


modules/pkg/paramscreator.go
```diff
@@ -58,6 +73,10 @@ func (r *run) ParamsCreator() (interface{}, error) {
 		switch checkType {
 		case "name":
 			p.PkgMatch.Matches = append(p.PkgMatch.Matches, checkValue)
+		case "version":
+			p.VerMatch = checkValue
+		case "notver":
```
jvehent: What about supporting a `!` operator like the one in the content parameter of the file module?

ameihm0912:
@jvehent yeah that makes sense, I'll change it to be similar

PR #234, opened 2016-05-31
[minor] reorganize api endpoints to satisfy my OCD-ness

jvehent:
This may or may not be a good idea.


jvehent:
Actually, don't merge this yet. The describe calls shouldn't have authentication at all.

ameihm0912:
r+

PR #243, opened 2016-06-09
[doc] update travisci to go1.6.2 and pg9.5

jvehent:


jvehent:
Close until travisci supports 9.5

PR #248, opened 2016-06-23
Enhancements to mig-agent-search

ameihm0912:
Rather than only supporting search via a targeting string, add a second mode that allows for agent queries that could include agents we wouldn't return during target evaluation. This can be used to list all agents known to the API.


ameihm0912:
@jvehent do you see any potential impact associated with the change in ce53035?

This is a small change cleaning up mig-agent-search with a few enhancements. The second commit changes DB.SearchAgents() to returns environment/tag details, where previously they were not there. I just want to make sure this will not introduce any performance issues I'm not considering.

jvehent:
r+

PR #250, opened 2016-07-05
[medium] change agent proxy behavior to prefer proxies

### LINKED ISSUE #249: Revisit proxy support and order of use
Current code we attempt to use configured proxies after a direct connection to the relay fails.

It may be better to try the proxies first if they are configured, and fall back to a direct connection if they fail, or potentially look into a configuration option to use only proxies for connectivity.

ameihm0912:
If any proxies have been configured, try those first before attempting
direct connections. This also adds specifying proxies in the agent
config file, where as before they were only compile time options in the
built-in configuration.

Resolves #249


ameihm0912:
@jvehent r?

jvehent:
r+

Issue #247, opened 2016-06-21
Validate X-FORWARDED-FOR contains what we expect in ip endpoint


ameihm0912:
In getIP() in the API, we look for an IP address in X-FORWARDED-FOR and return it if found. No validation occurs to confirm it is a valid IP, we could add a check here since the intent is for it to be used in the agent environment, and only return it if it's what we'd expect to be there.

claudijd:
Maybe also consider whether certain IPs in X-FORWARDED-FOR are acceptable too.  For instance, if someone provided X-Forwarded-For: 127.0.0.1.

ameihm0912:
Note this can actually result in a bug, but it's under a rare circumstance.

If agents are configured to connect via proxies, and the API is being offered without TLS (just HTTP), the append behavior for X-Forwarded-For will result in the comma delimited list of IPs being returned from the proxy path (if more than one are involved), which will panic the agent if it requests it's public IP.

Any API actually deployed should have TLS in front of it, in which case only a single X-Forwarded-For value should be present (added by the TLS termination) for a valid request.

@jvehent do you have any thoughts on if we should maintain support for running the API without TLS? If so, we would need some extra logic around how to handle the X-Forwarded-For header. It seems like a good idea given the standalone installer deploys it like this.

As an example, maybe an API configuration option that says when it should be preferred over the sockets peer address. For example when an agent is connecting through a proxy to a non-TLS API directly, we will not want to use X-Forwarded-For as the public address.

PR #251, opened 2016-07-05
Improve handling of X-Forwarded-For value in API

ameihm0912:
See #247 for details

Currently, if the API sees an X-Forwarded-For header, it will implicitly use this value over the peer address. This can cause issues in a scenario where the API is listening for connections directly on a non-SSL port, and an intermediate proxy is used by an agent. This results in the incorrect public IP being returned to the agent.

This adds a configuration option to the API to indicate if the API should honor the XFF value, or if it should use the peer address, which can be toggled depending on how it is deployed.

The default setting is off (do not use X-Forwarded-For) as the API alone operates in this manner, even though it is unlikely to be deployed like this, but this works with the standalone install.

### DIFF
--- a/conf/api.cfg.inc
+++ b/conf/api.cfg.inc
@@ -29,6 +29,16 @@
     #     |------<host>--------|<base>|--<endpoint>--|
     baseroute = "/api/v1"
 
+    # informs the api where it should obtain the clients public ip address
+    # from. the default if unset is "peer".
+    #
+    # use the X-Forwarded-For header, the trailing integer
+    # indicates an offset from the end of the list of addresses in
+    # x-forwarded-for to use as the client public ip:
+    #clientpublicip = x-forwarded-for:0
+    # use socket peer address:
+    #clientpublicip = peer
+
 [postgres]
     host = "127.0.0.1"
     port = 5432
--- a/doc/configuration.rst
+++ b/doc/configuration.rst
@@ -478,6 +478,16 @@ which is 192.168.1.150, on port 51664. The public endpoint of the api is
 		# ex: http://localhost:12345/api/v1/action/create/
 		#     |------<host>--------|<base>|--<endpoint>--|
 		baseroute = "/api/v1"
+                
+                # informs the api where it should obtain the clients public ip address
+                # from. the default if unset is "peer".
+                #
+                # use the X-Forwarded-For header, the trailing integer
+                # indicates an offset from the end of the list of addresses in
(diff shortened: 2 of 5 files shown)


conf/api.cfg.inc
```diff
@@ -29,6 +29,16 @@
     #     |------<host>--------|<base>|--<endpoint>--|
     baseroute = "/api/v1"
 
+    # informs the api where it should obtain the clients public ip address
+    # from. the default if unset is "peer".
+    #
+    # use the X-Forwarded-For header, the trailing integer
+    # indicates an offset from the end of the list of addresses in
+    # x-forwarded-for to use as the client public ip:
```
jvehent: Maybe add an example header with the corresponding offset?

doc/configuration.rst
```diff
@@ -509,6 +519,16 @@ Under the `[server]` section:
...
 * `host` is the public URL of the API, that clients will be connecting to
 * `baseroute` is the location of the base of the API, without the trailing slash.
+* `clientpublicip` tells where API where to get the clients public IP address.
+
+Ensure clientpublicip is set based on your environment. If clients are terminated
+directly on the API, peer can be used. If a load balancer or other device terminates
+connections from clients and adds the address to X-Forwarded-For, x-forwarded-for
+can be used. The integer trailing X-Forwarded-For specifies the offset from the end
+of the list of IPs in the header to use to extract the IP. For example,
+x-forwarded-for:0 would get the last IP in a list in that header, x-forwarded-for:1
```
jvehent: oh, here's my example :)

mig-api/api.go
```diff
@@ -295,20 +296,35 @@ func authenticateLoader(pass handler) handler {
...
+		if xff != "" {
+			xargs := strings.Split(xff, ",")
+			if ctx.Server.ClientPublicIPOffset >= len(xargs) {
+				ctx.Channels.Log <- mig.Log{Desc: "warning: requested X-Forwarded-For offset is not possible, not enough elements"}
+				useip = strings.Trim(xargs[0], " ")
+			} else {
+				useip = strings.Trim(xargs[(len(xargs)-1)-ctx.Server.ClientPublicIPOffset], " ")
+			}
+		} else {
+			ctx.Channels.Log <- mig.Log{Desc: "warning: API configured to use X-Forwarded-For but header not found"}.Info()
```
jvehent: if it's `warning`, it's not `Info()` ;)

PR #252, opened 2016-07-10
[major] hashed storage of api loader keys and format standardization

ameihm0912:
Changes to update the API to hash keys; this also standardizes the loader key format to a specific character set and length. In addition there are a few changes related to interaction for managing loader keys


ameihm0912:
@jvehent I may have a few additions or minor changes, but feel free to review if you have a moment

mig-api/loaderkey.go
```diff
@@ -0,0 +1,59 @@
...
+	hv := pbkdf2.Key([]byte(key), salt, 4096, 32, sha256.New)
+	return fmt.Sprintf("%x%x", salt, hv), nil
+}
+
+func hashAuthenticateLoader(lkey string) (ldr mig.LoaderEntry, err error) {
+	err = mig.ValidateLoaderPrefixAndKey(lkey)
+	if err != nil {
+		return
+	}
+	prefix := lkey[:mig.LoaderPrefixLength]
```
jvehent: Might want to be defensive here and check that `len(lkey) > mig.LoaderPrefixLength`

ameihm0912: I think this should be OK, the call to ValidateLoaderPrefixAndKey() right above includes a length check and will return an error if the key isn't the right format, including checking the length.

loader.go
```diff
@@ -16,20 +18,79 @@ import (
...
+// returns a random string consisting of alphanumeric characters of
+// length characters long
+func RandLoaderKeyString(length int) string {
+	ret := make([]byte, length)
+	lset := []byte("abcdefghijklmnopqrstuvwxyzABCDEFCHIJKLMNOPQRSTUVWXYZ0123456789")
+	r := mrand.New(mrand.NewSource(time.Now().UnixNano()))
+	for i := 0; i < len(ret); i++ {
+		ret[i] = lset[r.Int()%len(lset)]
+	}
+	return string(ret[:len(ret)])
```
jvehent: Or do a `strconv.FormatUint(random, 36)` like the GenID does.

loader.go
```diff
@@ -16,20 +18,79 @@ import (
...
+// The length of a loader key including the prefix
+const LoaderPrefixAndKeyLength = 40
+
+// The length of the prefix
+const LoaderPrefixLength = 8
+
+// Validate a loader key, returns nil if it is valid
 func ValidateLoaderKey(key string) error {
-	ok, err := regexp.MatchString("^[A-Za-z0-9]{1,256}$", key)
+	ok, err := regexp.MatchString("^[A-Za-z0-9]{32}$", key)
```
jvehent: Should that `32` use a constant like the rest?

loader.go
```diff
@@ -16,20 +18,79 @@ import (
...
 	if err != nil || !ok {
 		return errors.New("loader key format is invalid")
 	}
 	return nil
 }
+
+// Validate a loader prefix value, returns nil if it is valid
+func ValidateLoaderPrefix(prefix string) error {
+	ok, err := regexp.MatchString("^[A-Za-z0-9]{8}$", prefix)
```
jvehent: use `LoaderPrefixLength` instead of `8`

loader.go
```diff
@@ -16,20 +18,79 @@ import (
...
+	err := ValidateLoaderPrefix(pk[:8])
+	if err != nil {
+		return err
+	}
+	err = ValidateLoaderKey(pk[8:])
+	if err != nil {
+		return err
+	}
+	return nil
+}
```
jvehent: I'd prefer to have a separator between the loader prefix and key. Something simple like a `;` or `$` would do. It's not a big deal, just a bit easier for debugging when one needs to lookup a prefix in DB, for example.

mig-api/loaderkey.go
```diff
@@ -0,0 +1,59 @@
...
+	suppkey := lkey[mig.LoaderPrefixLength:]
+	id, hashkey, err := ctx.DB.GetLoaderIDAndKey(prefix)
+	if err != nil {
+		return
+	}
+	// Extract the 16 byte (32 char hex encoded) salt
+	salt, err := hex.DecodeString(hashkey[:32])
```
jvehent: Might want to test the length of hashkey, either here or in the DB lookup function.

client/mig-console/loader.go
```diff
@@ -110,13 +110,19 @@ r               refresh the loader entry (get latest version from database)
 			}
 			fmt.Printf("%v\n", string(jsonle))
 		case "key":
+			fmt.Println("New key component must be 32 alphanumeric characters long, or type 'generate' to generate one")
```
jvehent: I don't have an opinion on whether the length should be hardcoded or use constants, but you should use one or the other consistently, not both.

database/schema.sql
```diff
@@ -133,7 +133,8 @@ CREATE SEQUENCE loaders_id_seq START 1;
 CREATE TABLE loaders (
 	id            numeric NOT NULL DEFAULT nextval('loaders_id_seq'),
 	loadername    character varying(256) NOT NULL,
-	loaderkey     character varying(64) NOT NULL,
+	keyprefix     character varying(256) NOT NULL,
+	loaderkey     character varying(256) NOT NULL,
```
jvehent: here `loaderkey` is the hexadecimal of a salt+hash, correct?
as I mentioned on IRC, I think it may be cleaner to store salt and hash in their own column as bytes, without any encoding. We can always convert them on the flight using sql as needed.

jvehent:
I looked for the doc but couldn't find it. Maybe it got lost when you pushed the commits? :smiling_imp: 

r+, with comments

loader.go
ameihm0912: I'm not sure if this would work here. We want a consistent number of characters in the returned string. Depending on value of the randomly generated 64 bit number we could have anywhere from 1 to 13 characters in the resulting base 36 encoded output, meaning we will get different lengths in the return value.

ameihm0912: Note this is also why we have only a minimum here: https://github.com/mozilla/mig/blob/7751771158e06dc23220f533fbcdecf3c9475035/mig-agent/context.go#L317, since the resulting string length is variable.

PR #253, opened 2016-07-18
[medium] Add audit persistent module

arunk-s:
Add audit integration as a persistent module in mig


conf/mig-agent-conf.go.inc
```diff
@@ -156,3 +156,6 @@ MIIEpAIBAAKCAQEAvJQqCjE4I63S3kR9KV0EG9e/lX/bZxa/2QVvZGi9/Suj65nD
 ........
 RMSEpg+wuIVnKUi6KThiMKyXfZaTX7BDuR/ezE/JHs1TN5Hkw43TCQ==
 -----END RSA PRIVATE KEY-----`)
+
+//PERSISTENTMODULES contains module names and params that are to be started when the agent starts initially
+var PERSISTENTMODULES = [][]string {{"audit",`{"class":"parameters", "parameters":{"rulefilepath": "audit.rules.json", "outputsockets": ["/path/to/unix.sock"], "serveraddress": "mozdef_url"}}`}}
```
ameihm0912: We have a socket path option here and a server address; how does it decide which is used? is there a way to specify a subset of output modes to use or select a specific one?

mig-agent/agent.go
```diff
@@ -368,6 +377,24 @@ func runAgent(runOpt runtimeOptions) (err error) {
 	// service should simply be stopped.
 	exitReason = <-ctx.Channels.Terminate
 	ctx.Channels.Log <- mig.Log{Desc: fmt.Sprintf("Shutting down agent: '%v'", exitReason)}.Emerg()
+	// send stop messags to existing persistent modules
+	if len(runningPersistentMods) > 0 {
+		for modName := range runningPersistentMods {
+			msg := modules.Message{Class: modules.MsgClassStop, Parameters: nil}
+			response := make(chan []byte)
```
ameihm0912: Do we need to allocate a new channel for the response here or would we already have a channel open? It seems since we are in communication with the module for heartbeats and such as would already have something we could use here?

mig-agent/agent.go
```diff
@@ -457,6 +484,182 @@ func startRoutines(ctx *Context) (err error) {
...
+	if len(PERSISTENTMODULES) > 0 {
+		for _, moduleString := range PERSISTENTMODULES {
+			moduleName := moduleString[0]
+			ctx.Channels.Log <- mig.Log{Desc: fmt.Sprintf("Enabled persistent module %q", moduleName)}
+			//parse moduleparams from the config file
+			//module params are passed as JSON MsgClassParameters Message
+			moduleParams := moduleString[1]
+			ctx.Channels.Log <- mig.Log{Desc: fmt.Sprintf("Feeding message to persistent module %q", moduleParams)}
+			// feed params to peristent module
+			err = startPersistentModule(ctx, moduleName, moduleParams)
```
ameihm0912: Is there a way we could validate the persistent module parameters are correctly formatted here (moduleParams) before we attempt to start it?

mig-agent/agent.go
```diff
@@ -457,6 +484,182 @@ func startRoutines(ctx *Context) (err error) {
...
+	// open pipes for communicating with process stdin and stdout
+	stdinPipe, err := cmd.StdinPipe()
+	if err != nil {
+		return err
+	}
+	stdoutPipe, err := cmd.StdoutPipe()
+	if err != nil {
+		return err
+	}
+	stdoutScanner := bufio.NewScanner(stdoutPipe)
```
ameihm0912: We are probably OK here, one thing we might want to consider though is there is a fixed length to lines the scanner will correctly parse, from the docs:

When a scan stops, the reader may have advanced arbitrarily far past the last token. Programs that need more control over error handling or large tokens, or must run sequential scans on a reader, should use bufio.Reader instead. 

If we run into issues with long messages coming back we might want to switch to this.

conf/mig-agent-conf.go.inc
arunk-s: Currently, I don't decide on which option is to be used, as we are not writing to the unix socket yet and are dispatching events to url only. Though I think the behaviour can be modelled so that if both option are provided we write to both server and socket (does that make sense ?). However the option to specify an output mode and then providing the particular output address , seems a better option to me. (currently the code doesn't have this)

mig-agent/agent.go
arunk-s: Actually the channel denoted by `response` is to be used to notify back to the caller that wants to send message to the persistent module. Inspiration was taken from here https://github.com/jvehent/galilego/blob/master/main.go#L46 . The motivation behind is that we provide a method to send messages to the persistent module but at the same time the sender can continue to work and will not block to wait for response from the module.
It's used here for a similar purpose as we send stop message to the module on its supervisor channel and wait for their answer on the `response` channel for a specific period of time. In other scenarios, it would not make sense to use the heartbeat channel or the stdout channel of the module as: Multiple senders can send their requests to the module and we are already reading them in a goroutine but we want a way to notify them in a single mapped fashion. The channel is basically used as a signalling mechanism to notify the sender. Does it makes sense ?

mig-agent/agent.go
arunk-s: The module parameters are provided directly as a json string (defined in `modules.Message`) that should properly unmarshal into the module's parameter. Currently if any error happens while unmarshal the module will simply panic. We can modify the behaviour to try to decode the string, if any error occurs we return an error saying the params passed are wrong.

mig-agent/agent.go
arunk-s: Yeah, I agree. I'll mark it as a github issue.

jvehent:
@arunk-s : Have you had a chance to revisit this patch to use the new persistent module @ameihm0912 implemented?

arunk-s:
@jvehent Unfortunately not yet. I've gone through the implementation of persistent architecture but haven't started on modifying the audit module. I'll try to start on this at the earliest. I should be able to come up with something by this weekend.

ameihm0912:
Closing this PR as equivalent functionality was merged in #391

PR #254, opened 2016-08-08
[minor] add support for user defined target macros

ameihm0912:
This provides the ability for users to define target macros in ~/.migrc,
as an example rather than requiring -t "status='online'", given a
section such as:

[targets]
    macro = allonline:status='online'

in .migrc, you can then do -t allonline. The intended specification
format is macro = name:target string. Multiple "macro" items can be
specified in the targets section.

### DIFF
--- a/doc/configuration.rst
+++ b/doc/configuration.rst
@@ -670,6 +670,13 @@ you can reuse with your own values.
 	[gpg]
 		home = "/home/myuser/.gnupg/"
 		keyid = "E60892BB9BD89A69F759A1A0A3D652173B763E8F"
+        [targets]
+                macro = allonline:status='online'
+                macro = idleandonline:status='online' OR status='idle'
+
+The targets section is optional and provides the ability to specify
+short forms of your own targeting strings. In the example above, 
+`allonline` or `idleandonline` could be used as target arguments.
 
 Make sure have the dev library of readline installed (`readline-devel` on
 rhel/fedora or `libreadline-dev` on debian/ubuntu) and `go get` the binary from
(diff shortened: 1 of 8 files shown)


ameihm0912:
@jvehent r?

doc/configuration.rst
```diff
@@ -670,6 +670,13 @@ you can reuse with your own values.
 	[gpg]
 		home = "/home/myuser/.gnupg/"
 		keyid = "E60892BB9BD89A69F759A1A0A3D652173B763E8F"
+        [targets]
+                macro = allonline:status='online'
+                macro = idleandonline:status='online' OR status='idle'
+
+The targets section is optional and provides the ability to specify
+short forms of your own targeting strings. In the example above, 
+`allonline` or `idleandonline` could be used as target arguments.
```
jvehent: Could you add a note to the command_line and console docs too? I'm afraid this configuration doc is already too large for anyone to find anything in it.

doc/console.rst
```diff
@@ -203,7 +203,8 @@ Action parameters can be edited prior to launching it:
 
 	launcher> setname Test action that pings google.com
 
-* **settarget** sets the target of the action. The target is evaluated
+* **settarget** sets the target of the action. Targets can either be a targeting
+  string, or a macro if defined in migrc. The target is evaluated
```
jvehent: Ha, it's already here for the console :) Maybe add a small example of macro here as well?

jvehent:
That looks good. r+

ameihm0912:
Updated sample config to include some example macros

PR #256, opened 2016-08-11
[medium] add environment validation to loader authorization

### LINKED ISSUE #255: Restricting manifests to specific loader builds
The general workflow right now for a manifest fetch is as follows:

- The loader creates an agent environment data structure
- Loader provides it to API, if authenticated the API uses this environment to determine what manifest to return. This is done using a SQL query, the first manifest target string that matches the submitted loader environment wins.

Since this data is client provided, it is possible for a loader instance with an intentionally modified environment to fetch a manifest it shouldn't get. An example scenario might be if a manifest is created to target loaders with a specific tag, and a loader instance modifies its environment to have this tag set.

This doesn't really buy you much, but you could get a different key or AMQP credential to use to connect to the relay.

ameihm0912:
Adds ability to specify an expected environment or tag query string to
loader instances which are verified prior to manifest lookup.

Resolves #255

### DIFF
--- a/client/client.go
+++ b/client/client.go
@@ -595,6 +595,46 @@ func (cli Client) GetLoaderEntry(lid float64) (le mig.LoaderEntry, err error) {
 	return
 }
 
+// Change the expect fields of an existing loader entry
+func (cli Client) LoaderEntryExpect(le mig.LoaderEntry, eval string) (err error) {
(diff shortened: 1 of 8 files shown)


ameihm0912:
@jvehent r?

PR #257, opened 2016-08-12
[minor/bug] permit larger form sizes for manifest uploads

ameihm0912:
Set our own limit on body read size here since the file uploads can exceed the default.


jvehent:
r+

PR #258, opened 2016-08-15
[medium] more granular permissions for investigators

ameihm0912:
This removes the previous implementation where an investigator could
either be set to an admin or not, and expands this to per-endpoint
permissions that can be set for individual investigators.

### DIFF
--- a/mig-api/api.go
+++ b/mig-api/api.go
@@ -82,45 +82,45 @@ func main() {
 
 	// Investigator resources that require authentication
 	s.HandleFunc("/search",
-		authenticate(search, 0)).Methods("GET")
+		authenticate(search, mig.PermSearch)).Methods("GET")
 	s.HandleFunc("/action",
-		authenticate(getAction, 0)).Methods("GET")
+		authenticate(getAction, mig.PermAction)).Methods("GET")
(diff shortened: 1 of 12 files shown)


ameihm0912:
@jvehent r?

ameihm0912:
Updated with more fine-grained privileges per-endpoint

The UI component needs some expansion to support the granularity and currently only operates on categories/selection macros

mig-api/api.go
```diff
@@ -250,7 +238,10 @@ func authenticate(pass handler, adminRequired bool) handler {
 		if !ctx.Authentication.Enabled {
 			inv.Name = "authdisabled"
 			inv.ID = 0
-			inv.IsAdmin = true
+			inv.Permissions.DefaultSet()
+			inv.Permissions.ManifestSet()
+			inv.Permissions.LoaderSet()
+			inv.Permissions.AdminSet()
```
jvehent: Should it just be `0xFFFFFFFFFFFFFFFF` to catch future permissions as well?

ameihm0912: Yeah good point, we could do something like inv.Permissions.FromMask(0xffffffffffffffff)

PR #261, opened 2016-08-18
Add tests for loader signature validation

ameihm0912:
Adds tests for loader signature validation; this also resolves an issue related to duplicate signatures in the manifest.

### DIFF
--- a/mig-loader/loader_test.go
+++ b/mig-loader/loader_test.go
@@ -1,0 +1,428 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/.
+//
+// Contributor: Aaron Meihm ameihm@mozilla.com [:alm]
+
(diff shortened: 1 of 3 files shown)


mig-loader/loader_test.go
```diff
@@ -0,0 +1,428 @@
...
+}`
+
+func bootstrap() {
+	ctx.Channels.Log = make(chan mig.Log, 0)
+	go func() {
+		for {
+			_ = <-ctx.Channels.Log
+		}
+	}()
+}
```
jvehent: ❓ 🐰 
what does this do?

jvehent:
r+ except for that weird bootstrap function that I can't understand

mig-loader/loader_test.go
ameihm0912: checkManifestSignature in the test functions logs, so this just sets up a routine to drain the logging channel so it doesn't block

jvehent: Ha, that makes sense. May I suggest renaming it to `drainLogs` then?

PR #263, opened 2016-08-19
[minor] invert the logic to get the client hostname on Linux

gdestuynder:
First try to get the kernel hostname, if this fail or is localhost, try to get the filesystem hostname via
`hostname--fqdn` which basically lookups /etc/hosts /etc/hostname /etc/nsswitch.conf, etc.
Also, make this function never fail as having "localhost" as hostname is better than not starting at all (with
the previous logic it would still start with "localhost" anyway)


ameihm0912:
Moved to #265

PR #266, opened 2016-09-09
[minor] Create an API endpoint for agents to retrieve PGP public keys…

### LINKED ISSUE #240: Create an API endpoint for agents to retrieve PGP public keys
Rather than hardcoding long public keys into the agents, we should provide an api endpoint at `/publickey/<pgp_fingerprint>` that returns the armored public key that corresponds to the pgp fingerprint. Agent could then populate their local keyring using data from the API and only have the fingerprints in their local configuration.

sunnygkp10:
[minor] Create an API endpoint for agent to retrieve PGP public Key fixes #240

### DIFF
--- a/mig-api/api.go
+++ b/mig-api/api.go
@@ -18,6 +18,7 @@ import (
 	"github.com/gorilla/mux"
 	"github.com/jvehent/cljs"
 	"mig.ninja/mig"
+	"mig.ninja/mig/pgp"
 )
 
 var ctx Context
@@ -72,6 +73,7 @@ func main() {
 	// unauthenticated endpoints
 	s.HandleFunc("/heartbeat", getHeartbeat).Methods("GET")
 	s.HandleFunc("/ip", getIP).Methods("GET")
+	s.HandleFunc("/publickey/{pgp_fingerprint}", getPublicKey).Methods("GET")
 
 	// Loader manifest endpoints, use loader specific authentication on
 	// the request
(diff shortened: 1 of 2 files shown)


mig-api/api.go
```diff
@@ -438,6 +440,37 @@ func getIP(respWriter http.ResponseWriter, request *http.Request) {
...
+			respond(http.StatusInternalServerError, emsg, respWriter, request)
+		}
+		ctx.Channels.Log <- mig.Log{OpID: opid, Desc: "leaving getPublicKey()"}.Debug()
+	}()
+	vars := mux.Vars(request)
+	fp := vars["pgp_fingerprint"]
+
+	// retrieve the publickey
+	var inv mig.Investigator
+	if len(fp) > 0 {
```
ameihm0912: Should we just error out here if fp isn't valid? Right now in this case it looks like we would still call pgp.ArmorPubKey() later with the uninitialized investigator value, since that part of the function will still run.

ameihm0912:
@sunnygkp10 any chance you could update https://github.com/mozilla/mig/blob/master/doc/api.rst as well as part of this so we document the new API endpoint ?

mig-api/api.go
sunnygkp10: actually i was not sure  that what to do at this point so i need your suggestion. Although if fp isn't valid then it returns nothing but we can also add error message here.

sunnygkp10:
@ameihm0912 sure sir i can update .

mig-api/api.go
ameihm0912: You can probably just wrap something like https://github.com/mozilla/mig/blob/master/mig-api/investigator_endpoints.go#L59-L64 inside if fp == "" { }, and reply with No fingerprint specified in the error message.

mig-api/api.go
```diff
@@ -438,6 +440,37 @@ func getIP(respWriter http.ResponseWriter, request *http.Request) {
...
+		inv, err = ctx.DB.InvestigatorByFingerprint(fp)
+		if err != nil {
+			panic(err)
```
jvehent: If the error is `not found`, the API should return a 404, not a 503.

sunnygkp10:
@jvehent Sir, Please review the validation of responses.

mig-api/api.go
ameihm0912: We could probably just do if (fp != "") { grabFP(); } else { respond(400 error) } but otherwise looks good, r+

ameihm0912:
other then one noted nit which we can fix up after looks good, r+.

mig-api/api.go
sunnygkp10: So may i update that or its okay ?

sunnygkp10:
@ameihm0912 mentioned change is done, please have a look. :)

PR #267, opened 2016-09-13
[Minor]Clients should provide method to disable color terminal escape…

sunnygkp10:
I am providing a fix to this bug for review after that i will complete this whole task. 
I added a new struct Terminal in configuration and write method of that which is able to enable/disable color according to flag set in .migrc.
So if in future for writing colorful in terminal we will use the method of struct Terminal instead of direct fmt.Printf etc. 
There is also a argument which decide the color of text.

### DIFF
--- a/client/client.go
+++ b/client/client.go
@@ -46,10 +46,11 @@ type Client struct {
 
 // Configuration stores the live configuration and global parameters of a client
 type Configuration struct {
-	API     ApiConf    // location of the MIG API
-	Homedir string     // location of the user's home directory
-	GPG     GpgConf    // location of the user's secring
-	Targets TargetConf // Target macro specification
+	API      ApiConf      // location of the MIG API
+	Homedir  string       // location of the user's home directory
+	GPG      GpgConf      // location of the user's secring
+	Targets  TargetConf   // Target macro specification
+	Terminal TerminalConf // Terminal configuration
 }
 
 type ApiConf struct {
@@ -65,6 +66,28 @@ type GpgConf struct {
 type TargetConf struct {
 	Macro  []string
 	macros map[string]string
+}
+
+type TerminalConf struct {
+	Color bool
+}
+
+//Used for color print in terminal base on color flag
+//in configuration file
+func (clr TerminalConf) Printf(c string, s string, a ...interface{}) {
+	if clr.Color {
+		fmt.Printf("\x1b["+c+"m"+s+"\x1b[0m", a...)
+	} else {
+		fmt.Printf(s, a...)
+	}
+}
+
+func (clr TerminalConf) Fprintf(w io.Writer, c string, s string, a ...interface{}) {
+	if clr.Color {
+		fmt.Fprintf(w, "\x1b["+c+"m"+s+"\x1b[0m", a...)
+	} else {
+		fmt.Fprintf(w, s, a...)
+	}
 }
 
 // Used by the macro parser to add processed target macros to the client
@@ -266,6 +289,7 @@ func MakeConfiguration(file string) (err error) {
 	}
 	fmt.Fprintf(fd, "[api]\n\turl = \"%s\"\n", cfg.API.URL)
 	fmt.Fprintf(fd, "[gpg]\n\thome = \"%s\"\n\tkeyid = \"%s\"\n", cfg.GPG.Home, cfg.GPG.KeyID)
+	fmt.Fprintf(fd, "[TerminalConf]\n\tcolor = off\n")
 	fd.Close()
 	fmt.Println("MIG client configuration generated at", file)
 	return
--- a/client/mig-console/console.go
+++ b/client/mig-console/console.go
@@ -53,19 +53,6 @@ func main() {
 	}
 	defer out.Close()
 
-	fmt.Fprintf(out, "\x1b[32;1m"+banner+"\x1b[0m")
-
-	// append a space after completion
-	readline.CompletionAppendChar = 0x20
-	// load history
-	historyfile := homedir + "/.mig_history"
-	fi, err := os.Stat(historyfile)
-	if err == nil && fi.Size() > 0 {
-		err = readline.LoadHistory(historyfile)
-		if err != nil {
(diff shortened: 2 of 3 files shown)


sunnygkp10:
@jvehent @ameihm0912 Please review the implementation.

client/client.go
```diff
@@ -46,10 +46,11 @@ type Client struct {
...
 type Configuration struct {
-	API     ApiConf    // location of the MIG API
-	Homedir string     // location of the user's home directory
-	GPG     GpgConf    // location of the user's secring
-	Targets TargetConf // Target macro specification
+	API      ApiConf      // location of the MIG API
+	Homedir  string       // location of the user's home directory
+	GPG      GpgConf      // location of the user's secring
+	Targets  TargetConf   // Target macro specification
+	Terminal TerminalConf // Terminal configuration
```
ameihm0912: Seems like this would always just remain a single option, maybe we don't need a new section for it in that case and just include it in the main configuration?

client/client.go
```diff
@@ -266,6 +289,7 @@ func MakeConfiguration(file string) (err error) {
 	}
 	fmt.Fprintf(fd, "[api]\n\turl = \"%s\"\n", cfg.API.URL)
 	fmt.Fprintf(fd, "[gpg]\n\thome = \"%s\"\n\tkeyid = \"%s\"\n", cfg.GPG.Home, cfg.GPG.KeyID)
+	fmt.Fprintf(fd, "[TerminalConf]\n\tcolor = off\n")
```
ameihm0912: Would be good to use lower case section/key names here to keep consistent with the rest of the file

client/client.go
```diff
@@ -67,6 +68,28 @@ type TargetConf struct {
...
 
+type TerminalConf struct {
+	Color bool
+}
+
+//Used for color print in terminal base on color flag
+//in configuration file
+func (clr TerminalConf) Printf(c string, s string, a ...interface{}) {
+	if clr.Color {
+		fmt.Printf("\x1b["+c+"m"+s+"\x1b[0m", a...)
```
ameihm0912: off hand looking at this, does this assume the entire line will utilize the same terminal color/attributes? does this cause problems in place where we have various highlighting currently on the same line.

given the current code is it possible for us to just wrap the output string in a function that strips the control codes if colorized output has been disabled by the user? That would cover off all the requires cases

client/client.go
sunnygkp10: It is for future scope may be we need to add more option.

client/client.go
sunnygkp10: first of all if we have various highlighting in same line then we will call function for separately for all color and the and there is color attribute in function which consist of two things color+ text decoration . so it is not a problem . 
and yes if color flag in configuration is off then it will print the simple text.

client/client.go
sunnygkp10: But the above method is so lengthy. Sorry for your time. I want to write code which work as
```
if (flag == off )
  remove (\x1b[__;_m and \x1b[0m)
```
from entire code before building binary. Currently i have no idea how to do that but i will do.
any suggestion ?

sunnygkp10: It seems like a lengthy way. I am thinking of another function which parse the text which is going to print and remove all the escape sequence from it if color flag is off. 
````
func checkColor(s string) string {
	if color == on {
		return s
	} else {
		s = strings.Replace(s, "\x1b[0m", "", -1)
		var buffer bytes.Buffer
		color_code_flag := 0
		for i := 0; i < len(s); i++ {
			if i+4 <= len(s) && string(s[i]) == "[" && (string(s[i+3]) == ";" || string(s[i+4]) == ";") {
				color_code_flag = 1
			} else if string(s[i]) == "m" {
				color_code_flag = 0
			}
			if color_code_flag == 0 {
				buffer.WriteString(string(s[i]))
			}
		}
		return buffer.String()
	}

}
```
which is call as 
	`fmt.Println(checkColor("\x1b[31;1m+------\x1b[0m"))`
@ameihm0912 please give feedback and apologies for taking time.

sunnygkp10:
It seems like a lengthy way. I am thinking of another function which parse the text which is going to print and remove all the escape sequence from it if color flag is off. 
````
func checkColor(s string) string {
	if color == on {
		return s
	} else {
		s = strings.Replace(s, "\x1b[0m", "", -1)
		var buffer bytes.Buffer
		color_code_flag := 0
		for i := 0; i < len(s); i++ {
			if i+4 <= len(s) && string(s[i]) == "[" && (string(s[i+3]) == ";" || string(s[i+4]) == ";") {
				color_code_flag = 1
			} else if string(s[i]) == "m" {
				color_code_flag = 0
			}
			if color_code_flag == 0 {
				buffer.WriteString(string(s[i]))
			}
		}
		return buffer.String()
	}

}
```
which is call as 
	`fmt.Println(checkColor("\x1b[31;1m+------\x1b[0m"))`
@ameihm0912 please give feedback and apologies for taking time.

ameihm0912:
@sunnygkp10 yeah what you suggest there is along the lines of what I was thinking. I'm wondering if there is a nicer way we could do it though, instead of wrapping the arguments each time, like maybe just a generalized client output function (e.g., client.Out, client.OutErr); and we just call that vs. calling the fmt related functions, and handle stripping the codes there.

sunnygkp10:
@ameihm0912 There are some package for handling color in golang. for example [fatih/color](https://github.com/fatih/color). They did the same thing as you said but problem is that they are only for print function in terminal and we use color in input function (such as [readline](https://github.com/mozilla/mig/blob/master/client/mig-console/console.go#L106) etc) also.
if you guide me i can write a small library for color which satisfy our all requirement and then we will rewrite our existing color output according to that.
or we can use above `checkcolor()` function.

jvehent:
@sunnygkp10 : Have you had a chance to revisit this PR?

sunnygkp10:
@jvehent yes, i want to complete it but currently i am busy with my course work .

ameihm0912:
Closing due to inactivity, feel free to reopen if still in progress

PR #268, opened 2016-09-14
[minor] support passphrase input via env

ameihm0912:
Not intended for use in the general case, but permits using various mig
command line tools in an automated manner more easily.


ameihm0912:
@jvehent r?

PR #272, opened 2016-09-25
[medium/bug] Prevent one investigator from signing multiple times

jvehent:
This patch fixes a security issue in the ACL verification code
that allows a single investigator to sign an action multiple
times in order to meet the threshold required to run the action.

Given the lack of support for multiple signatures in the client
code until recently, and the fact the investigator still has to
be authorized to run the module anyway, this is considered MEDIUM.

r? @ameihm0912

### DIFF
--- a/acl.go
+++ b/acl.go
@@ -28,13 +28,21 @@ func verifyPermission(operation Operation, permName string, perm Permission, fin
 		return fmt.Errorf("Invalid permission '%s'. Must require at least 1 signature, has %d",
 			permName, perm[permName].MinimumWeight)
 	}
+	var seenFp []string
 	signaturesWeight := 0
 	for _, fp := range fingerprints {
+		// if the same key is used to sign multiple times, return an error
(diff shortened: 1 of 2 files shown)


jvehent:
Ha. I missed that in your patch. I've had this patch sitting in the distributedacls for a year and only remembered about it when I saw your changes. I don't mind dropping this PR, I don't think we need it in both places. Bigger question is: should this check live in action.go or acl.go ?

ameihm0912:
Seems OK either way, since the action signatures always call the ACL verification code. I'm OK with whichever you prefer, we can take it out of the action verification, the one advantage of that I guess is it happens sooner prior to validating against the ACL.

jvehent:
The only argument I have for putting it in acl.go is `VerifyPermission` can currently be gamed by a single investigator. I seems somewhat odd to allow for it to happen.

ameihm0912:
Sounds good to me, r+, we can remove the other code from the action verification

PR #275, opened 2016-10-09
Persistent module support

ameihm0912:
Adds support for persistent modules.

Persistent modules can be used to run background tasks in the agent as a separate process which is managed and kept alive by the agent. The modules run and can keep state over their lifetime, and can be used to perform investigative related tasks across time or execute other continuous functions.

Persistent modules can be queried in the same manner as regular modules using mig-cmd or mig-console. The Run() entry point is still used, but rather than return a result string this function instead interfaces directly with the running module over a domain socket. This removes any need to multiplex requests to the module in the agent itself, runModuleDirectly() is still used for a request which interfaces with the domain socket the persistent module maintains.

The example module periodically logs the current time to the agent, and echos any string sent to it by a client.

### DIFF
--- a/modules/modules.go
+++ b/modules/modules.go
@@ -30,10 +30,15 @@ import (
 var ModuleRunDir string
 
 // Message defines the input messages received by modules.
+//
+// All messages will have Class and Parameters set. PersistSock is used in a case
+// where a parameters message is being sent for a persistent module. In this case,
+// PersistSock will contain the socket specification the module has registered as
+// listening on.
 type Message struct {
-	Class       MessageClass `json:"class"`                // represent the type of message being passed to the module
-	Parameters  interface{}  `json:"parameters,omitempty"` // for `parameters` class, this interface contains the module parameters
-	PersistSock string       `json:"persistsock,omitempty"`
+	Class       MessageClass `json:"class"`                 // represent the type of message being passed to the module
+	Parameters  interface{}  `json:"parameters,omitempty"`  // for `parameters` class, this interface contains the module parameters
+	PersistSock string       `json:"persistsock,omitempty"` // Persistent module socket path if required
 }
 
 type MessageClass string
@@ -46,10 +51,12 @@ const (
 	MsgClassRegister   MessageClass = "register"
 )
 
+// Parameter format expected for a log message
 type LogParams struct {
 	Message string `json:"message"`
 }
 
+// Parameter format expected for a register message
 type RegParams struct {
 	SockPath string `json:"sockpath"`
 }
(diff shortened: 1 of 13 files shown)


ameihm0912:
@jvehent @arunk-s @gdestuynder any prelim review appreciated. This is intended to provide a framework for integration of the audit module.

modules/modules.go
```diff
@@ -184,6 +236,106 @@ func WatchForStop(r io.Reader, stopChan *chan bool) error {
...
+			WriteOutput([]byte(resp), conn)
+			err = conn.Close()
+			if err != nil {
+				panic(err)
+			}
+		}()
+	}
+}
+
+func SendPersistRequest(p interface{}, modname string) string {
```
jvehent: Should return an error

modules/modules.go
```diff
@@ -184,6 +236,106 @@ func WatchForStop(r io.Reader, stopChan *chan bool) error {
...
+			}
+		}
+
+		if failed {
+			break
+		}
+	}
+}
+
+func HandlePersistRequest(l net.Listener, f func(interface{}) string) {
```
jvehent: Should return an error

modules/modules.go
```diff
@@ -106,6 +123,21 @@ func MakeMessage(class MessageClass, params interface{}, comp bool) (rawMsg []by
...
+	defer func() {
+		if e := recover(); e != nil {
+			err = fmt.Errorf("Failed to make module log message: %v", e)
+		}
+	}()
+	param := LogParams{Message: fmt.Sprintf(f, args...)}
+	msg := Message{Class: MsgClassLog, Parameters: param}
+	rawMsg, err = json.Marshal(&msg)
+	if err != nil {
+		panic(err)
```
jvehent: Since you only have one error check, maybe remove the panic recover entirely?

modules/modules.go
```diff
@@ -184,6 +236,106 @@ func WatchForStop(r io.Reader, stopChan *chan bool) error {
...
+	}
+	rb, err := ioutil.ReadAll(conn)
+	if err != nil {
+		panic(err)
+	}
+	return string(rb)
+}
+
+func PersistSockPath(modname string) string {
```
robmurtha: Consider registering and mapping the name to both the net and address parameters so that when looking up by name both the net and address parameters can be returned and passed to Listen and Dial for supporting tcp connections on Windows (if needed).

ameihm0912:
@arunk-s in that scenario (running mig-agent -m <module>); if the agent is running and the persist module is running, it will operate as expected (read params on stdin, write results on stdout). if no agent is running and you run the module directly with -m, you'd get a results message with an error indicating the socket was not available.

I think this _should_ be ok. From a testing perspective, you can run mig-agent -P persistmod, and then mig-agent -m persistmod at the same time to query the persistent module independent of the agent running (for testing purposes, etc).

modules/modules.go
ameihm0912: Yeah good point, my hope was the net library was portable enough it perhaps implemented this as a named pipe and we'd be ok, but that doesn't appear to be the case. I'll see if i can add some code around this case so we can drop something in that will work for Windows.

ameihm0912:
@robmurtha the change in 192860a here should give us more flexibility on Windows

PR #279, opened 2016-10-28
[minor] additional signing for osx to be compatible with sierra

### LINKED ISSUE #278: console messages (500,000 of them)
Getting a bazillion of these messages (looks like my console has a limit of 500k messages.)
Looking at ps auxwww output, it appears the process is generally sleeping. So it doesn't seem to be causing any issues other than this message being repeated. So i can't say when this started except maybe when I upgraded to Sierra on the 16th.

**/kernel  (/System/Library/Extensions/AppleMobileFileIntegrity,kext/Contents/MacOS/AppleMobileFileIntegrity)**
- default   15:37:16.296978 -0500   kernel  AMFI: allowing exception handler for 'mig-agent' (708) because the process is unsigned.
- default   15:37:16.297053 -0500   kernel  AMFI: allowing exception handler for 'mig-agent' (708) because the process is unsigned.

**ps output:**
- 10114  mig-agent    0.0       00:00.00 5      0    24    1316K  0B     0B     10109 10109 sleeping *0[1]             0.00000 0.00000    0    1148        73      29          14
- 10109  mig-loader   0.0       00:00.28 7      1    50    10M    0B     0B     10109 1     sleeping *0[1]             0.00000 0.00000    0    5940        215     148         50

ameihm0912:
We want to make sure signatures are applied to the compiled binaries in
addition to the loader package.

Resolves #278

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -14,6 +14,25 @@ else
 	BINSUFFIX := ""
 endif
 
+# These variables control signature operations used when building various
+# targets on OSX.
+#
+# OSXPROCSIGID if set will result in the specified identity being used to
+# sign the mig-agent and mig-loader binaries when built on OSX. If empty,
+# the compiled binaries will not be signed.
+#
+# OSXPACKSIGID if set will result in the specified identify being used to
+# sign the mig-loader package (osx-loader-pkg). If empty the .pkg will not
+# be signed.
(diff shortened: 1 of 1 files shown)


Phrozyn:
+r

PR #285, opened 2016-11-04
[minor] windows fix agent path

### LINKED ISSUE #282: Agent fails to create `C:\Program Files\mig\` on Windows 10
The default path on windows is hardcoded in env_windows.go as `C:\Program Files\mig\`, the agent cannot create the path and fails to start on Windows 10.

robmurtha:
use C:\mig instead of C:\Program Files\mig\
closes #282

### DIFF
--- a/mig-agent/agentcontext/env_windows.go
+++ b/mig-agent/agentcontext/env_windows.go
@@ -106,5 +106,5 @@ exit:
 }
 
 func GetRunDir() string {
-	return `C:\Program Files\mig\`
+	return `C:\mig\`
 }


ameihm0912:
Seems reasonable to me. I'm not familiar enough with that platform to really know what the best place is to put this. @jvehent do you have any thoughts on impact here to anything before we merge this?

jvehent:
Isn't there a `%PROGRAMFILES%` variable we could use?

robmurtha:
There is a whole host of directories referenced in environment variables, many do not exist. The issue contains a reference to where to put installed program data. This basically just unblocks us from running on windows until we can sort out a proper install location.

Issue #229, opened 2016-05-10
Return the list of enabled modules in the agent heartbeat


jvehent:
We should store the list of enabled modules in the heartbeat the agent sends back to the scheduler. This would allow targeting agents that support a given module, and in the future list agents that did not manage to initialize a given persistent module.

ameihm0912:
Closed via #295

PR #295, opened 2016-12-07
[Minor] List of agent's enabled modules added to heartbeat according to issue #229

Caiyeon:
Adds functionality as per issue #229

### DIFF
--- a/agent.go
+++ b/agent.go
@@ -46,6 +46,7 @@ type AgentEnv struct {
 	Addresses []string    `json:"addresses,omitempty"`
 	PublicIP  string      `json:"publicip,omitempty"`
 	AWS       AgentEnvAWS `json:"aws,omitempty"`
+	Modules   []string    `json:"modules,omitempty"`
 }
 
 // Stores AWS specific agent environment values
--- a/mig-agent/agent.go
+++ b/mig-agent/agent.go
@@ -871,7 +871,7 @@ func sendResults(ctx *Context, result mig.Command) (err error) {
 	return
 }
 
-// hearbeat will send heartbeats messages to the scheduler at regular intervals
+// heartbeat will send heartbeats messages to the scheduler at regular intervals
 // and also store that heartbeat on disc
 func heartbeat(ctx *Context) (err error) {
 	// loop forever
--- a/mig-agent/context.go
+++ b/mig-agent/context.go
@@ -27,6 +27,7 @@ import (
 	"github.com/jvehent/service-go"
 	"github.com/streadway/amqp"
 	"mig.ninja/mig"
+	"mig.ninja/mig/modules"
 	"mig.ninja/mig/mig-agent/agentcontext"
 )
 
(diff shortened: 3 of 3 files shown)


agent.go
```diff
@@ -33,6 +33,7 @@ type Agent struct {
 	Authorized      bool        `json:"authorized,omitempty"`
 	Env             AgentEnv    `json:"environment,omitempty"`
 	Tags            interface{} `json:"tags,omitempty"`
+	Modules         []string    `json:"modules,omitempty"`
```
ameihm0912: I'm wondering if we should include this information in the agent's environment element here (Env), vs. putting it in a new part of the heartbeat. The advantage of putting it in the environment is we get this information inserted into the database without any changes there. If we want to put it outside the environment, I think we should modify the scheduler code to also insert this information into the agents table.

See for example here: https://github.com/mozilla/mig/blob/master/mig-scheduler/agents_queues.go#L125 where we process the incoming heartbeat, and here: https://github.com/mozilla/mig/blob/master/database/agents.go#L97

One of the goals with this is to be able to target agents running specific modules via -t to the command line and so on, so we will need this information in the database.

Caiyeon:
I've relocated the module updates to mig-agent/context.go; This seems to be closer to the ideal location. I have a couple of concerns.

Is it possible that a module starts up with the agent, but fails later on?

At the moment everything in the AgentEnv comes from AgentContext. AgentContext's principal is to only gather information regarding the system, rather than the agent itself. Thus, by placing the module update in this location, the "env has changed" return value is questioned. If the module availability changes, has the environment changed? AgentContext has not changed, but AgentEnv has changed.

ameihm0912:
> Is it possible that a module starts up with the agent, but fails later on?

It's possible, but the agent will keep trying to restart it. Right now, we don't have any sort of concept of a "disabled" module state. So, modules.Available will be static and shouldn't change given a build of the agent.

I see what you mean about AgentContext. As you mentioned this is really meant for grabbing information about the system that can change. Some of the data in ctx.Agent.Env can change based on the AgentContext, but some of it stays the same when the agent runs (like Env.Proxies).

I'm wondering if we can treat this like we do the Proxy related values, and just set it on startup? For example, this is where the proxy values get set, https://github.com/mozilla/mig/blob/master/mig-agent/context.go#L493-L494.

Could we just set env.Modules somewhere in Init() for example?

Caiyeon:
Placed it in Init(). I originally placed it in the update function thinking it wasn't static.

While this allows targeting agents with specific modules, it does not guarantee that the agent's specified module is fully functional. But I suppose with the current structure this is as close as it gets.

ameihm0912:
@Caiyeon thanks, this looks good to go

Can we just amend the commit according to https://github.com/mozilla/mig/blob/master/CONTRIBUTING.md?

We add a tag to the front of the commit message indicating the scope of the change, so for example in this case I'd suggest [minor]. Once we have that I think it's ready to merge.

PR #297, opened 2016-12-10
[medium] persistent module config file support

ameihm0912:
Adds the ability for persistent modules to read a configuration file
that can be used to control their behavior.

In some cases, we may want a persistent module to be configurable. A new
configuration variable PERSISTCONFIGDIR has been added which indicates a
directory the agent will look in for persistent module configs. When the
agent starts a persistent module, it will determine if modulename.cfg
exists in the config directory; if so the agent will load the config and
send it to the newly executed module, which it will read on stdin. The
module can then read this configuration, and configure itself as needed
before it begins it's primary execution functions.

Modules specify the configuration format within the module code itself,
so each module that makes use of this can have it's own configuration
parameters.

### DIFF
--- a/conf/mig-agent-conf.go.inc
+++ b/conf/mig-agent-conf.go.inc
@@ -49,7 +49,7 @@ var SPAWNPERSISTENT = true
 // XXX This should be improved to take into account Windows paths, but at this
 // time persistent module support is not available on Windows. The agent will
 // attempt to locate a configuration using the module name, e.g., modulename.cfg.
-var PERSISTCONFIGDIR = "/etc/mig"
+var MODULECONFIGDIR = "/etc/mig"
 
 // how often the agent will refresh its environment. if 0 agent
 // will only update environment at initialization.
--- a/conf/mig-agent-conf.go.inc
+++ b/conf/mig-agent-conf.go.inc
@@ -5,7 +5,7 @@
 // Contributor: Julien Vehent jvehent@mozilla.com [:ulfr]
 package main
 
-import(
+import (
 	"mig.ninja/mig"
 	"time"
 )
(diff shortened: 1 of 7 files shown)


ameihm0912:
@arunk-s we should be able to use this to specify a config for the audit module if you wanna give it a try. I updated examplepersist to illustrate how.

arunk-s:
@ameihm0912 This looks cool. I'll try this out with the audit module. :)

conf/mig-agent-conf.go.inc
```diff
@@ -43,13 +43,21 @@ var EXTRAPRIVACYMODE = false
 // disabled at run-time using a config option or command line flag
 var SPAWNPERSISTENT = true
 
+// The directory the agent will look for persistent module configuration files
+// in.
+//
+// XXX This should be improved to take into account Windows paths, but at this
+// time persistent module support is not available on Windows. The agent will
+// attempt to locate a configuration using the module name, e.g., modulename.cfg.
+var PERSISTCONFIGDIR = "/etc/mig"
```
jvehent: Does this need to be persistent module specific, or could it be used by any module? If so, should it be called `MODULESCONFDIR` instead?

mig-agent/persist.go
```diff
@@ -64,6 +67,19 @@ func (p *persistModuleRegister) remove(modname string) {
...
+// module is started. If the configuration file cannot be loaded, just return the
+// config struct for the module uninitialized.
+func getPersistConfig(modname string) (ret interface{}) {
+	cfg := modules.Available[modname].NewRun().(modules.PersistRunner).PersistModConfig()
+	confpath := path.Join(PERSISTCONFIGDIR, modname+".cfg")
+	// An error here isn't fatal, we just continue with cfg as is
+	gcfg.ReadFileInto(cfg, confpath)
+	ret = cfg
+	return
+}
```
jvehent: Two thoughts?
1. Should this code belong to the `modules` package instead, just in case other components need to use it?
2. Should it enforce the `gcfg` format, or leave it up to the module to unmarshal the config into a format of its choice? I'm concerned `gcfg` might limit us a bit too much, but maybe not.

modules/examplepersist/examplepersist.go
```diff
@@ -63,12 +63,35 @@ var logChan chan string
 // returning and the module exiting.
 var handlerErrChan chan error
 
+// When the agent sends the persistent module it's configuration, it will come in via
+// the config channel as a JSON byte slice so we can unmarshal it into our configuration
+var configChan chan []byte
```
jvehent: I think passing the configuration through the call to `Run()` would be easier, or are you using a channel in case we want to change the configuration without reloading the module?

conf/mig-agent-conf.go.inc
ameihm0912: Makes sense, yea

PR #299, opened 2016-12-16
[medium/bug] resolve bufio issues under high agent/module comms

ameihm0912:
When a message between a module and the agent is read by either,
ReadInput is used which uses buffered IO to read a single line. A bug
existed here where a new bufio Reader was being allocated each time.

Under normal circumstances there would be no issue, since a single line
is generally sent to the module from the agent and read. However, in a
case where a large number of messages were being exchanged between the
agent and the module (e.g., persistent modules) it was possible the
bufio reader could have buffered more than one line. Since a new buffer
was being allocated each time, this resulted in loss of partial message
data, and usually ended up generating JSON unmarshaling errors since
part of the JSON structure was missing.

This was resolved by introducing a new ModuleReader and ModuleWriter
type that is used to read/write between the agent and modules.
ModuleReader allocates the buffered reader on creation once, so we only
ever have the single buffered reader for the lifetime of the module
execution.

### DIFF
--- a/client/mig/main.go
+++ b/client/mig/main.go
@@ -244,7 +244,7 @@ func main() {
 		if err != nil {
 			panic(err)
 		}
-		out := run.(modules.Runner).Run(bytes.NewBuffer(msg))
+		out := run.(modules.Runner).Run(modules.NewModuleReader(bytes.NewBuffer(msg)))
 		if len(out) == 0 {
 			panic("got empty results, run failed")
 		}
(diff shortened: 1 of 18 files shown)


ameihm0912:
For reference, the issue was here: https://github.com/mozilla/mig/blob/d85cc3960ae9dcaca86ccd4a440d60ae96ea56e9/modules/modules.go#L191

We create reader each time there; if there is a lot of data in the underlying reader r, the next time we come back to ReadInput() a new reader is created, and we just lose all the buffered data we had previously.

PR #300, opened 2016-12-19
addition of fswatch module

ameihm0912:
fswatch is a simple file system monitoring module that takes advantage
of the persistent module framework in MIG to do continuous file
integrity scans of specific paths on the file system. When changes are
detected, alerts are generated and appear in the agent's log file.

Currently has a built-in configuration for monitoring that is platform specific, that can be overridden using a module configuration file.


modules/fswatch/fswatch.go
```diff
@@ -0,0 +1,215 @@
...
+	return
+}
+
+var logChan chan string
+var handlerErrChan chan error
+var configChan chan []byte
+
+var alertChan chan FSWatchAlert
+
+// Alert severities
```
gdestuynder: just wondering if we could/should use standard levels

csjperon:
My name is csjp and I approve this commit.

modules/fswatch/fswatch.go
```diff
@@ -0,0 +1,215 @@
...
+	alertChan <- FSWatchAlert{
+		Severity: sev,
+		Alert:    fmt.Sprintf(f, a...),
+	}
+}
+
+// An alert generated by the fswatch module
+type FSWatchAlert struct {
+	Severity int    `json:"severity"` // ALERT_CRITICAL, etc...
+	Alert    string `json:"alert"`    // alert text
```
jvehent: nit on naming: the module is already called `fswatch` so no need to repeat it in the struct name. Also, the `Alert` field could be more descriptive, like `Message` or `Body`.
```go
type Alert struct {
    Severity int
    Message string
}
```

modules/fswatch/fswatch_profile_linux.go
```diff
@@ -0,0 +1,22 @@
...
+		{"/var/spool/cron", []fsWatchObject{}},
+		{"/bin", []fsWatchObject{}},
+		{"/sbin", []fsWatchObject{}},
+		{"/usr/bin", []fsWatchObject{}},
+		{"/usr/sbin", []fsWatchObject{}},
+		{"/etc", []fsWatchObject{}},
+		{"/etc/init.d", []fsWatchObject{}},
+		{"/etc/systemd/system", []fsWatchObject{}},
+	},
+}
```
jvehent: That's very broad for a default profile. I'd prefer to be a lot more specific and give administrators the choice of going with a broad coverage, to limit the noise when enabling this with a default conf.

modules/fswatch/paramscreator.go
```diff
@@ -0,0 +1,22 @@
...
+package fswatch /* import "mig.ninja/mig/modules/fswatch" */
+
+import (
+	"fmt"
+)
+
+func printHelp(isCmd bool) {
+	fmt.Printf(`Query parameters
+----------------
+This module has no parameters.
```
jvehent: Should it even implement these interfaces? `ParamsParser` and `ParamsCreator` are really for command line tools, and I don't see any use of those here.

modules/fswatch/watch.go
```diff
@@ -0,0 +1,256 @@
...
+// contain a single entry for the file path.
+type fsWatchProfileEntry struct {
+	path    string
+	objects []fsWatchObject
+}
+
+// Collect all entries in a directory indicated by a profile entry, note that we do
+// not do recursion here and only read the first directory level
+func (f *fsWatchProfileEntry) collectDirectory() error {
+	dirents, err := ioutil.ReadDir(f.path)
```
jvehent: `filepath.Walk` could be used to traverse the directory tree recursively.

modules/fswatch/fswatch.go
jvehent: Syslog standard levels?

ameihm0912:
@jvehent:

> I'm surprised that you're not using built-in file monitoring facilities, like inotify

Yeah, that is definitely a better approach. My initial thought here was to implement it in a platform independent manner, and then add some platform specific optimizations on top, but talking to @gdestuynder yesterday he pointed out https://github.com/fsnotify/fsnotify which may be pretty suitable for this, so I'm going to see what that looks like here.

> Maybe a file system cache would be better so we don't need those in memory?

I'm sort of hesitant to start writing to the file system here, but it may end up being required at some point so can keep it in mind and we can add this if needed.

> On a more strategic level, I'm questioning the ultimate value of this module compared 
> to the audit module. Don't they do the same thing? Are we getting more context on a change
> here than we would by watching audit events on specific files? I can see how being
> cross-platform is a plus here, but maybe the audit module could be extended to be
> cross-platform as well?

Correct me if I am wrong here, but I think doing this with audit would be possible but not ideal. In addition to watching for file writes, we'd need to also build a heuristic to understand the implications of various other system calls that modify the file system (e.g., rename, link) and based on these determine if a file may have changed, which would work but I don't think is a more optimal solution for this.

modules/fswatch/paramscreator.go
ameihm0912: Eventually I'd like to return stats (e.g., number of files monitored, module resource utilization, etc); right now the module is query-able but it just returns the status if its running.

jvehent:
> Correct me if I am wrong here, but I think doing this with audit would be possible but not ideal. In addition to watching for file writes, we'd need to also build a heuristic to understand the implications of various other system calls that modify the file system (e.g., rename, link) and based on these determine if a file may have changed, which would work but I don't think is a more optimal solution for this.

@gdestuynder would know more

modules/fswatch/fswatch.go
gdestuynder: mozilla standard levels

gdestuynder:
Audit can do the same thing, yes. It's not multi-platform though which is what I assume(d) is the reason for not using it.

The various notify frameworks should work about as well as audit for this very use case though, with a caveat: it most likely won't have as much metadata associated as with auditd.

For reference, here are some things audited for by audit on many systems (w means write, a means attribute change)

`-w /usr/bin -p wa -k binaries
-w /bin -p wa -k binaries
-w /usr/sbin -p wa -k binaries
-w /sbin -p wa -k binaries
-w /usr/local/bin -p wa -k binaries
-w /usr/local/sbin -p wa -k binaries
-w /etc/audit/ -p wa -k audit
-w /etc/audisp/ -p wa -k audit
-w /etc/sysconfig/auditd -p wa -k audit
-w /etc/libaudit.conf -p wa -k audit
-w /etc/rsyslog.conf -p wa -k syslog
-w /etc/rsyslog-ng/ -p wa -k syslog
-w /etc/syslog.conf -p wa -k syslog
-w /etc/syslog-ng.conf -p wa -k syslog
-w /etc/syslog-ng/ -p wa -k syslog
-w /data/hekad/hekad.toml -p wa -k heka
-w /etc/hekad/hekad.toml -p wa -k heka
-w /etc/cron.allow -p wa -k cron
-w /etc/cron.deny -p wa -k cron
-w /etc/cron.d/ -p wa -k cron
-w /etc/cron.daily/ -p wa -k cron
-w /etc/cron.hourly/ -p wa -k cron
-w /etc/cron.monthly/ -p wa -k cron
-w /etc/cron.weekly/ -p wa -k cron
-w /etc/crontab -p wa -k cron
-w /var/spool/cron/root -p wa -k cron
-w /etc/rc.d/init.d/ -p wa -k init
-w /sbin/init -p wa -k init
-w /etc/inittab -p wa -k init
-w /etc/systemd -p wa -k init
-w /etc/pam.d -p wa -k pam
-w /etc/security -p wa -k pam
-w /lib/security -p wa -k pam
-w /etc/sshd -p wa -k sshd
-w /etc/group -p wa -k user
-w /etc/passwd -p wa -k user
-w /etc/gshadow -p wa -k user
-w /etc/shadow -p wa -k user
-w /etc/security/opasswd -p wa -k user
-w /etc/sudoers -p wa -k user
-w /etc/localtime -p wa -k time-change
`

modules/fswatch/fswatch_profile_linux.go
```diff
@@ -6,17 +6,17 @@
...
+var localFsWatchProfile = profile{
+	entries: []profileEntry{
+		{path: "/boot", recursive: false},
+		{path: "/etc/cron.d", recursive: false},
+		{path: "/var/spool/cron", recursive: false},
+		{path: "/bin", recursive: false},
+		{path: "/sbin", recursive: false},
+		{path: "/usr/bin", recursive: false},
+		{path: "/usr/sbin", recursive: false},
+		{path: "/etc", recursive: true},
```
gdestuynder: I suspect this covers /etc/systemd and /etc/init.d and /etc/cron.d with recursive=true
also, /lib /usr/lib /lib64 /usr/lib64 are similar to /bin /usr/bin etc (i.e. might want to watch them too)

PR #303, opened 2016-12-21
[minor] change the way AGTCONF and LOADERCONF are handled in Makefile

ameihm0912:
When building the mig-agent or mig-loader targets and specifying AGTCONF
or LOADERCONF, configuration.go in the relevant directory is replaced by
the referenced configuration. If AGTCONF or LOADERCONF is not
overridden, the configuration is not replaced.

This can lead to inconsistent build results, as by building for example
the mig-agent target a second time with no AGTCONF, the replaced
configuration.go will be retained.

This changes this behavior, so configuration.go is always replaced with
the configuration files in conf/, unless the Makefile configuration
variable has been overridden.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -58,8 +58,8 @@ endif
 PREFIX		:= /usr/local/
 DESTDIR		:= /
 BINDIR		:= bin/$(OS)/$(ARCH)
-AGTCONF		:= mig-agent/configuration.go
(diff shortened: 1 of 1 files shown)


ameihm0912:
Disregard, will resubmit as we could solve a similar problem with available_modules

PR #304, opened 2016-12-22
Improvements to Makefile

ameihm0912:
Cleans up some elements of the Makefile, and adds some new targets to make the output of builds more predictable


ameihm0912:
@jvehent should be fixed now

Issue #236, opened 2016-06-06
Don't require ACLs in MIG Agent


jvehent:
ACLs should be optional to allow simpler deployment to let any investigator with a configured pgp fingerprint to send actions.
In code terms, that means putting `VerifyACL` in `mig-agent/acl.go` behind a configuration flag.

jboyer2012:
Hello.
I'd love to jump into this project and this looks like a good point. If this is still desired, can you give me a quick overview/example of how configuration flags are handled in MIG? Even if it's just another file I'd like to see an example of a configuration flag in action.

ameihm0912:
@jboyer2012 the configuration flags for the agent you can find here:

These can be considered "defaults", and are compiled into the agent:
https://github.com/mozilla/mig/blob/master/mig-agent/configuration.go

Then, if a configuration file exists, the defaults are overridden using the configuration file in here:
https://github.com/mozilla/mig/blob/master/mig-agent/config.go

So what you'd probably want to do is add a new option, VERIFYACLS or something along those lines, which can be true or false, defaulting to true. If false, we wouldn't require an ACL configuration, just a valid signature, which we'd verify using the public key list in the agent.

The verification starts here:
https://github.com/mozilla/mig/blob/master/mig-agent/acl.go#L52

Then eventually ends up here:
https://github.com/mozilla/mig/blob/master/action.go#L362-L369

That code block does verification of signatures, with fingerprints eventually containing a list of all the valid key fingerprints used to sign the action, so I think we could add something after that to make sure fingerprints is not empty and if our configuration flag is set to not require ACL verification we just return.

For a simple example of how you might use a bool config flag in the agent, here's one: https://github.com/mozilla/mig/blob/master/mig-agent/agent.go#L443

jboyer2012:
@ameihm0912 Thanks for the great overview! Can this be easily tested with a standalone installation?

ameihm0912:
Yep, no problem implementing this with the standalone installation

jboyer2012:
One more question, is there a page with standards/guidelines for contributing?

ameihm0912:
Should all be at https://github.com/mozilla/mig/blob/master/CONTRIBUTING.md

jboyer2012:
[...]

PR #314, opened 2017-01-31
[minor] Add VERIFYACLs global to bypass verification of ACLs if set to false. Implements #236

jboyer2012:


action.go
```diff
@@ -343,7 +343,7 @@ func (a Action) String() (str string, err error) {
 // permission is found, the default one `default` is used.
 // The first permission that is found to apply to an operation, but
 // doesn't allow the operation to run, will fail the verification globally
-func (a Action) VerifyACL(acl ACL, keyring io.Reader) (err error) {
+func (a Action) VerifyACL(acl ACL, keyring io.Reader, verifyACLs bool) (err error) {
```
ameihm0912: I'm wondering if we should use something other than verifyACLs for the flag that is a bit clearer.

Perhaps onlyVerifyPubkey or something like that?

action.go
```diff
@@ -368,6 +368,14 @@ func (a Action) VerifyACL(acl ACL, keyring io.Reader) (err error) {
 		fingerprints = append(fingerprints, fp)
 	}
 
+	if len(fingerprints) == 0 {
+		return errors.New("No valid fingerprints found.")
+	}
+
+	if !verifyACLs {
```
ameihm0912: If we invert the logic above (require true to bypass ACL check) we'd need to also switch this around.

ameihm0912:
Can we rebase against master for this branch too ?

action.go
jboyer2012: Do you mean throughout the entire application (in the configuration.go and config.go files) or just the variable passed in to the VerifyACL function?

ameihm0912: Was thinking just here, so if onlyPubkey (or whatever we decide to call it) is true, we only verify the key, but if it's false (default) the ACL code is run too

jboyer2012: I just committed [9ad057e](https://github.com/jboyer2012/mig/commit/9ad057e22962def33050b75aea3f40b7aeaf7519) for these changes. I decided to change the flag everywhere. After thinking about it, I feel it is clearer in its current form. I also reworded the comments to make more explicit what the flag does. Let me know if you like the changes.

jboyer2012: Sorry just realized the last commit I failed to update the actual call to the VerifyAcls function. [d177bd8](https://github.com/jboyer2012/mig/commit/d177bd86dc821d70fe5a1a570d0b478f8bb0dda8) fixed it.

ameihm0912:
Looks good, the travis build is failing as the stand alone install needs to be updated but I can fix that up. Thanks for grabbing this.

PR #315, opened 2017-02-02
Better correlation between loader/agent instances and action logging

ameihm0912:
Adds an option to the scheduler to log a summary of each command dispatched to an agent. This summary will include the name of the loader instance associated with the agent, if one is present.

To do this, agent ID (e.g., .migagtid) related functionality was moved into agentcontext so it could be used by both the loader and the agent. The loader now sends the agent ID with its requests, which will be the same ID the agent is using for queue location to allow correlation between the instances.

### DIFF
--- a/database/agents.go
+++ b/database/agents.go
@@ -248,7 +248,8 @@ func (db *DB) ActiveAgentsByTarget(target string) (agents []mig.Agent, err error
 	}
 	rows, err := txn.Query(fmt.Sprintf(`SELECT DISTINCT ON (queueloc) id, name, queueloc,
 		version, pid, starttime, destructiontime, heartbeattime, refreshtime, status,
-		mode, environment, tags
+		mode, environment, tags,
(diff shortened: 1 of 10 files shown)


database/agents.go
```diff
@@ -248,7 +248,8 @@ func (db *DB) ActiveAgentsByTarget(target string) (agents []mig.Agent, err error
 	}
 	rows, err := txn.Query(fmt.Sprintf(`SELECT DISTINCT ON (queueloc) id, name, queueloc,
 		version, pid, starttime, destructiontime, heartbeattime, refreshtime, status,
-		mode, environment, tags
+		mode, environment, tags,
+		(SELECT loadername FROM loaders WHERE loaders.queueloc = agents.queueloc LIMIT 1)
```
jvehent: This is only executed when scheduling new actions, so I think it's fine if it isn't the fastest query in the world. Still,you seem to be retrieving data for the solve purpose of logging, and I can't help but wonder if that shouldn't just be put in a separate query that's part of the logging function `logAgentAction`.

PR #316, opened 2017-02-02
pinentry runPass doesn't disable echo

gcoxmoz:
https://github.com/camlistore/camlistore/issues/897

I was hoping to fix this upstream and then have this get pulled down, but there's more hoops than I can reasonably jump through, not being a Go programmer.

Running mig in a minimal environment leaves my password being echoed back on the screen, which is bad.  If you can improve this or get it upstream, have at it.

### DIFF
--- a/pgp/pinentry/pinentry.go
+++ b/pgp/pinentry/pinentry.go
@@ -120,8 +120,14 @@ func (r *Request) GetPIN() (pin string, outerr error) {
 
 func runPass(bin string, args ...string) {
 	cmd := exec.Command(bin, args...)
+	cmd.Stdin  = os.Stdin
 	cmd.Stdout = os.Stdout
-	cmd.Run()
+	cmd.Stderr = os.Stderr
+	err := cmd.Run()
+	if err != nil {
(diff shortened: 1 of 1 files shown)


ameihm0912:
@gcoxmoz thanks for this

Yeah the environment you describe isn't really a typical environment the client would run in, but the fix makes sense. The Travis builds are failing due to some errors, do you want to fix this up or would you prefer I grab it?

I'm not sure what sort of environment you are setting up, but note one other option you have too is there is a means to pass a passphrase in as an environment variable to the client (https://github.com/mozilla/mig/blob/master/client/client.go#L186-L198) which might simplify what you are doing too if it's suitable for your requirements.

gcoxmoz:
Please, have at it.  I can't `go` way out of a paper bag, so when upstream wanted me to go through hurdles and their IDE and etc... here I am.

I'm working on making mig-the-client build in docker (so I can not have all its weird requirements cluttering my laptop).  Knowing the ENV variable is there is cool, and I'll work on dropping that in as an option, but, I think having this work is a good step forward.

ameihm0912:
Fixes merged in #317

PR #317, opened 2017-02-03
[minor] update pgp getPINNaïve so stty -echo is successful

ameihm0912:
See #316 for additional details


gcoxmoz:
Verified, working.  Thanks!

PR #321, opened 2017-03-07
[minor] build standalone instance into Docker container

jvehent:


ameihm0912:
@jvehent looks like the build is failing as docker is missing

jvehent:
@ameihm0912: ok to merge?

PR #323, opened 2017-03-12
[minor] Don't create a client conf when running with -t local

jvehent:


client/mig/main.go
```diff
@@ -299,6 +282,24 @@ func main() {
 	}
 
 readytolaunch:
+
```
ameihm0912: I think if we move this here, it will cause problems with https://github.com/mozilla/mig/blob/da684c5c5f981fd73c10ced5a192eaf51cf0f2c0/client/mig/main.go#L278 (depends on a read configuration structure being present in the client type)

jvehent:
Good catch! I moved it further up to avoid the issue.

PR #328, opened 2017-03-20
[minor] Add an option to print agent settings, fixes #280

### LINKED ISSUE #280: Add an option to print agent settings
It would be nice to have a way to provide a flag to mig-agent, where it just outputs it's current settings it would run with. This would provide a means to see how the built-in configuration is setup, and how the config file options override the built-in configuration.

Currently there is no easy method to see the configuration of the agent after it has been compiled.

vbmade2000:
@ameihm0912 please take a look.


ameihm0912:
@vbmade2000 looks good, thanks!

PR #333, opened 2017-03-27
[minor] include recent actions in agent statistics output

ameihm0912:
Agent tracks some basic stats on the last 15 actions it has received and provides details on these via the status socket. This includes if the action was processed or not.

Currently this is limited to in-memory stats which would be lost on restart, but the intent is it can be adapted to make use of local storage for persistence if needed.

### DIFF
--- a/mig-agent/stats.go
+++ b/mig-agent/stats.go
@@ -15,6 +15,7 @@ package main
 
 import (
 	"mig.ninja/mig"
+	"strings"
 	"sync"
 	"time"
 )
@@ -51,12 +52,27 @@ type agentStatsAction struct {
 	Time     string
 	Name     string
 	Accepted string
+	Modules  string
 }
 
 // Add data to an agentStatsAction based on action a
 func (s *agentStatsAction) importAction(a mig.Action, accepted bool) error {
 	s.Time = time.Now().UTC().Format(time.RFC3339)
 	s.Name = a.Name
+	fl := make([]string, 0)
+	for _, x := range a.Operations {
+		found := false
+		for _, y := range fl {
+			if y == strings.ToLower(x.Module) {
+				found = true
+			}
+		}
+		if found {
+			continue
+		}
+		fl = append(fl, strings.ToLower(x.Module))
+	}
+	s.Modules = strings.Join(fl, ", ")
 	if accepted {
 		s.Accepted = "Accepted"
 	} else {
--- a/mig-agent/stats.go
+++ b/mig-agent/stats.go
@@ -19,8 +19,6 @@ import (
 	"time"
 )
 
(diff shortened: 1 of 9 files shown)


mig-agent/stats.go
```diff
@@ -0,0 +1,73 @@
...
+// The information here can be presented by querying the agent on its status
+// socket.
+
+import (
+	"mig.ninja/mig"
+	"sync"
+	"time"
+)
+
+const maxStatActions = 15 // Keep stats for last X number of actions
```
jvehent: This should be an agent config parameter.

mig-agent/stats.go
```diff
@@ -0,0 +1,73 @@
...
+// Stats we keep for an action processed by the agent
+type agentStatsAction struct {
+	Time     string
+	Name     string
+	Accepted string
+}
+
+// Add data to an agentStatsAction based on action a
+func (s *agentStatsAction) importAction(a mig.Action, accepted bool) error {
+	s.Time = time.Now().UTC().Format("Jan 2 15:04:05 2006")
```
jvehent: Maybe use RFC3339 for the time format?

mig-agent/stats.go
```diff
@@ -0,0 +1,73 @@
...
+	}
+	return nil
+}
+
+// Initialize agent statistics
+func initAgentStats(ctx Context) (newctx Context, err error) {
+	newctx = ctx
+	ctx.Stats = agentStats{}
+	ctx.Channels.Log <- mig.Log{Desc: "Agent statistics management initialized"}.Info()
+	return
```
jvehent: It doesn't look like you need the initialization. Do you?

mig-agent/stats.go
```diff
@@ -0,0 +1,73 @@
...
+	}
+	s.Actions = append(s.Actions, ns)
+	return nil
+}
+
+// Stats we keep for an action processed by the agent
+type agentStatsAction struct {
+	Time     string
+	Name     string
+	Accepted string
```
jvehent: investigator name?
list of modules invoked?

mig-agent/stats.go
ameihm0912: Not really right now, this is more of a placeholder to be used to initialize persistent statistics (e.g., if we use sqlite or something) on disk in the future

mig-agent/stats.go
ameihm0912: Added modules

Investigator names is a bit trickier, that information is not sent with the action to the agents (appears this is only populated when actions are retrieved from the API). We'd have to pull this out of the key ring and there isn't a nice way to return that right now from the ACL verification code.

Issue #209, opened 2016-04-16
YARA module


jvehent:
YARA is a client tool for scanning file system for IOCs. [Go-Yara](https://github.com/hillu/go-yara) implements binding to the yara C library that we could ship as a module via cgo.

On the client side, we should load YARA files into MIG actions, possibly compressed, the same way we load scribe files.

note: this is not an easy one.

jvehent:
@hillu: do you know if the Yara library can be statically compiled into a Go binary, or if the only way to ship it is the dynamically link to libyara?

hillu:
* Julien Vehent [:ulfr]:

> @hillu: do you know if the Yara library can be statically compiled
> into a Go binary, or if the only way to ship it is the dynamically
> link to libyara?

It's possible to link statically. Use something like

    go build -x --ldflags '-extldflags "-static"'

However, on Linux the resulting binary would still be dynamically linked
against GNU libc which may not be what you want for MiG. Using libmusl
instead is a solid option, though.

It is also possible to cross-build a static libyara for Windows on Linux
(and presumably MacOS) hosts using MinGW.

Cheers,
-Hilko

ameihm0912:
I've got a working implementation of this for Linux in #335 which uses go-yara and looks good.

Doing the same for the OSX builds seems rather troublesome.

Out of the box with the changes, just trying to compile it for OSX:

```
GOOS=darwin GOARCH=amd64 GO15VENDOREXPERIMENT=1 go build  -tags netgo -o bin/darwin/amd64/mig-agent-20170329-0.e3b4219.dev"" -ldflags "-X mig.ninja/mig.Version=20170329-0.e3b4219.dev  -extldflags '-static '" mig.ninja/mig/mig-agent
# mig.ninja/mig/vendor/github.com/hillu/go-yara
vendor/github.com/hillu/go-yara/compiler.go:15:10: fatal error: 'yara.h' file not found
#include <yara.h>
         ^
1 error generated.
make: *** [mig-agent] Error 2
```

Add a few flags so we can find things

```
GOOS=darwin GOARCH=amd64 GO15VENDOREXPERIMENT=1 go build  -tags netgo -o bin/darwin/amd64/mig-agent-20170329-0.e3b4219.dev"" -ldflags "-X mig.ninja/mig.Version=20170329-0.e3b4219.dev  -extldflags '-static '" mig.ninja/mig/mig-agent
# mig.ninja/mig/mig-agent
[...]

PR #335, opened 2017-03-29
Yara module

ameihm0912:
This implements a yara module as described in #209.

It is currently limited to scanning file system objects, and has been tested on Linux and Darwin agents.

Makes use of the file module to identify candidate files to scan using supplied yara rules.

Note that in order to build this and link the yara library in statically to the agent, a few changes were made to the Makefile.

WITHYARA=yes, this further tweaks linker flags to ensure a binary is produced with required yara dependencies built in. This requires a few libraries be present on the build system to produce the agent and is documented in the module documentation.


ameihm0912:
@jvehent this should be good for some review, feel free to take a quick look whenever you have a chance. Thanks

Issue #239, opened 2016-06-06
Replace API X-PGP-Authorization with basic tokens


jvehent:
X-PGP-Authorization was a good bad idea. While very secure because the server never sees the private key, it's also non-standard and thus hard to implement in clients that don't directly use the MIG codebase.
The motivation behind X-PGP-Authorization was to avoid managing another set of keys when investigators already had to manage a PGP key to sign actions. However, we now have several client types that do not have PGP keys (loader, runner, some webui dashboard in the future) and requiring x-pgp-authorization just makes things difficult.

A simple API Token (oauth bearer token? jwt? hawk?) would suffice and be more standard.

jvehent:
JWT may be a better fit.

jvehent:
[Stripe](https://stripe.com/docs/api#authentication) uses HTTP Basic Auth on their API, and that doesn't seem like a terrible idea. It's way simpler than any other mechanism, which helps integration. The downside being we need to store API keys server side, when JWT is a pubkey mechanism.

ameihm0912:
We currently have a couple authentication functions being used, authenticate() and authenticateLoader() https://github.com/mozilla/mig/blob/master/mig-api/api.go#L231-L326

One potential option is we adapt the approach used in authenticateLoader to the authenticate function

This would require the addition of properties such as the key prefix/hash to the investigator's entry in the database as well

jvehent:
So we need to rip out `authenticate()` entirely, rename `authenticateLoader()` into `authenticate()` and use it for all endpoints. Then we can create a new DB column in the investigator table where we store the API tokens of clients.

ameihm0912:
Here's one example of how we are currently hashing authorization strings for the loader https://github.com/mozilla/mig/blob/master/mig-api/loaderkey.go

jvehent:
Here's how the loader key and salt are stored in DB: https://github.com/mozilla/mig/blob/master/database/schema.sql#L137-L138

PR #339, opened 2017-04-11
Support API key authentication for mig API

ameihm0912:
Adds support for using a standard API key where required to access the MIG API. This essentially adds an alternative to using X-PGPAUTHORIZATION which may be more easily utilized for integration of external applications with the API.

Note: this changes nothing related to action signing and interrogating agents, and is solely utilized for simplified access to functions the API provides itself.

### DIFF
--- a/client/client.go
+++ b/client/client.go
@@ -1134,7 +1134,7 @@ func (cli Client) PostInvestigatorStatus(iid float64, newstatus string) (err err
 	return
 }
 
-// PostInvestigatorAPIKey is used to either enable or disable API key based access
+// PostInvestigatorAPIKeyStatus is used to either enable or disable API key based access
 // to the MIG API for an investigator. API key based access to the API can be used in
 // place of X-PGPAUTHORIZATION API authentication.
 //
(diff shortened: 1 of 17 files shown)


ameihm0912:
Also see #239

client/client.go
```diff
@@ -1124,6 +1134,54 @@ func (cli Client) PostInvestigatorStatus(iid float64, newstatus string) (err err
...
 
+// PostInvestigatorAPIKey is used to either enable or disable API key based access
+// to the MIG API for an investigator. API key based access to the API can be used in
+// place of X-PGPAUTHORIZATION API authentication.
+//
+// If an API key is being set, the returned investigator APIKey value will contain the
+// assigned key. newstatus should be set to either 'active' or 'disabled'. If a key already
+// exists for an investigator, calling this with a status of 'active' will cause the existing
+// key to be replaced.
+func (cli Client) PostInvestigatorAPIKey(iid float64, newstatus string) (inv mig.Investigator, err error) {
```
jvehent: I wonder if posting an api key is something we want, versus `GET /investigator/apikey` that would return a randomly generated one. Thoughts?

ameihm0912: This function is sort of poorly named; it does result in a randomly generated API key being returned right now, but it makes use of the existing /investigator/update endpoint which expects a POST. I will rename this to something more obvious.

PR #345, opened 2017-04-19
[minor] Adding option to set agent tags in external config file

rezen:
Apologies in advance, my go experience is limited at this point so any feedback is constructive!

Currently tags can only be set in the build of the agent binary. Portions of the agents configuration can be set in an external file, so I just added the ability to set tags in that file. The tags are in essence configured as a list of key value pairs separated by colon. At init the tags are parsed and set. Tags set externally are merged with the `TAGS` set in the agent and can be accessed using the new method, `GetTags()`

Tags can be set in this fashion.

```
[agent]
    ; Tag the instance with attributes to filter investigation scope
    tags = "operator:Ops_Emea"
    tags = "role:web"
```

### DIFF
--- a/conf/mig-agent.cfg.inc
+++ b/conf/mig-agent.cfg.inc
@@ -54,6 +54,11 @@
     ; if true, only the investigator's public key is verified on actions and not ACLs.
     onlyVerifyPubKey = false
 
+    ; Tag the instance with attributes to filter investigation scope
+    tags = "operator:Ops_Emea"
+    tags = "role:web"
+
+
 [stats]
     ; number of previous actions the agent should keep a summary for. the summary can
     ; be viewed over the agent stat socket. 0 to disable.
--- a/mig-agent/config.go
+++ b/mig-agent/config.go
@@ -36,6 +36,7 @@ type config struct {
 		PersistConfigDir string
 		ExtraPrivacyMode bool
 		OnlyVerifyPubKey bool
+		Tags             []string
 	}
 	Stats struct {
 		MaxActions int
@@ -167,6 +168,9 @@ type globals struct {
 
 	// Add the private client key below.
 	agentKey []byte
+
+	// Provides some meta data about agents and lets you target specific ones
+	tags map[string]string
 }
 
 func newGlobals() *globals {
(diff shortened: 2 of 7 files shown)


mig-agent/config.go
```diff
@@ -167,9 +183,30 @@ type globals struct {
...
+	for tag, val := range structs.Map(TAGS) {
+		field, ok := inspectTags.FieldByName(tag)
+
+		if ok {
+			_, tag = stringPair(string(field.Tag))
+			tag = strings.Replace(tag, `"`, "", -1)
+		}
+		tags[tag] = fmt.Sprintf("%v", val)
+	}
+	return tags
```
ameihm0912: I'm wondering if there is a way we could do this without needing to use reflect and structs?

Like for example, maybe we just say TAGS has to include a Runtime element that is defined as a slice of key/value strings. Then when the agent starts up, we just add anything in the config file into that slice which we will know exists already.

TAGS right now is just an arbitrary struct, but historically we've never really used it much for anything other than key/value pairs. Maybe we should just change the TAGS type to be a map an avoid the conversion, and it would make it easier to work with?

conf/mig-agent.cfg.inc
```diff
@@ -54,15 +54,20 @@
...
     maxactions = 15
 
-[certs]
-    ca  = "/path/to/ca/cert"
-    cert= "/path/to/client/cert"
-    key = "/path/to/private/key"
+; [certs]
+;    ca  = "/path/to/ca/cert"
+;    cert= "/path/to/client/cert"
+;    key = "/path/to/private/key"
```
ameihm0912: Not sure if the intent was to comment these out?

mig-agent/config.go
rezen: I'm all in favor of converting the TAGS to a map! I will test that out and make sure it works out ok!

conf/mig-agent.cfg.inc
rezen: That was intentional, I copy over the config in `tools/standalone_install.sh`. I could make the standalone_install comment out those lines instead of editing the template.

mig-agent/config.go
ameihm0912: @jvehent does this sound reasonable to you? I'm not sure what the original intent around TAGS being an arbitrary struct was. Do you see any issue if we just convert this so TAGS is assumed to be map[string]string, since generally that is how we are using it across the board anyway?

jvehent: Yeah, I agree defining it as a `map[string]string` is a good idea. Make sure to update the doc though.

rezen: Ok, changing the type will definitely make things easier, I'll be able to remove some of my workaround.

ameihm0912:
@rezen noticed the additional commit, this one ready for review?

rezen:
Yes, thank you!

mig-agent/config.go
```diff
@@ -167,6 +168,9 @@ type globals struct {
 
 	// Add the private client key below.
 	agentKey []byte
+
+	// Lets you target specific agents
```
ameihm0912: Small comment, can be used beyond just targeting (e.g., the tags just generally describe some properties of the agent, but can be used for targeting too).

ameihm0912:
@rezen generally looks good, added one minor comment but a couple others:

Could we gofmt the code? That is preferred for merge to keep things consistent (https://github.com/mozilla/mig/blob/master/CONTRIBUTING.md).

Also, can we take the changes to the configuration file and standalone install out? The standalone install agent uses a built-in configuration, tests are good but we should find a different way at some point as this will likely cause problems there.

rezen:
Ah yes forgot to run gofmt! Updated with the requested changes.

ameihm0912:
@rezen apologies for the delays on this; travis is failing here but I believe I see what's going on. Give me a bit to resolve that part of things and will merge.

rezen:
Checking in on this

PR #357, opened 2017-06-08
[minor] include actual pgp error on token sign failure

ameihm0912:


client/client.go
```diff
@@ -140,8 +140,8 @@ func NewClient(conf Configuration, version string) (cli Client, err error) {
 	// try to make a signed token, just to check that we can access the private key
 	_, err = cli.MakeSignedToken()
 	if err != nil {
-		err = fmt.Errorf("failed to generate a security token using key %s from %s\n",
-			conf.GPG.KeyID, conf.GPG.Home+"/secring.gpg")
+		err = fmt.Errorf("failed to generate a security token using key %v from %v: %v\n",
```
jvehent: unsure if it matters, but why chance the `%s` into `%v` ?

ameihm0912: Just figured while I'm here, seems more consistent with the way the rest of the code is currently

PR #364, opened 2017-07-11
[easy/bug] Check length of config.API.URL, fixes #362

### LINKED ISSUE #362: ReadConfiguration runtime error if file is unreadable
In client ReadConfiguration, a runtime error is generated if the configuration file is not readable, this should return a proper error

ReadConfiguration() -> runtime error: index out of range

tydavis:


ameihm0912:
@tydavis could we amend this commit so it's prefixed with [minor/bug] per https://github.com/mozilla/mig/blob/master/CONTRIBUTING.md ?

ameihm0912:
@tydavis disregard, after some discussion we are going to remove this prefix requirement; this is all good.

PR #365, opened 2017-07-11
[easy/up-for-grabs] Add additional line for unknown investigator

### LINKED ISSUE #235: Clarify error messages related to action signature validation failure
Currently when an agent receives an action with a signature it cannot validate we end up with an error message like:

parseCommands() -> checkActionAuthorization() -> action ACL verification failed: Failed to retrieve fingerprint from signatures: GetFingerprintFromSignature() -> openpgp: signature made by unknown entity

It might be worth adding a follow up message stating the action was rejected as it was sent by an unknown investigator or something along those lines.

tydavis:
Fixes #235

### DIFF
--- a/mig-agent/agent.go
+++ b/mig-agent/agent.go
@@ -513,7 +513,12 @@ func startRoutines(ctx *Context) (err error) {
 		for msg := range ctx.Channels.NewCommand {
 			err = parseCommands(ctx, msg)
 			if err != nil {
-				log := mig.Log{Desc: fmt.Sprintf("%v", err)}.Err()
+				var log mig.Log
+				if strings.Contains(err.Error(), "signature made by unknown entity") {
+					log = mig.Log{Desc: fmt.Sprintf("%v \n Action rejected -- sent by unknown investigator.", err)}.Err()
+				} else {
+					log = mig.Log{Desc: fmt.Sprintf("%v", err)}.Err()
+				}
 				ctx.Channels.Log <- log
 			}
 		}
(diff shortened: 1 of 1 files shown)


mig-agent/agent.go
```diff
@@ -513,7 +513,12 @@ func startRoutines(ctx *Context) (err error) {
 		for msg := range ctx.Channels.NewCommand {
 			err = parseCommands(ctx, msg)
 			if err != nil {
-				log := mig.Log{Desc: fmt.Sprintf("%v", err)}.Err()
+				var log mig.Log
+				if strings.Contains(err.Error(), "signature made by unknown entity") {
+					log = mig.Log{Desc: fmt.Sprintf("%v \n Action rejected -- sent by unknown investigator.", err)}.Err()
```
ameihm0912: @tydavis is this line feed in here because we are bundling two log lines up in a single message?

I'm wondering if instead of doing this, we should just send the "Action rejected" log message after sending the first one which contains the unknown entity error.

The reason being there are a couple of log output modes (e.g., syslog, windows event log) and I'm not sure what the behavior of these will be with multiline input.

mig-agent/agent.go
```diff
@@ -991,4 +996,5 @@ func printConfigSettings() {
 	fmt.Println("HEARTBEATFREQ     : ", HEARTBEATFREQ)
 	fmt.Println("MODULETIMEOUT     : ", MODULETIMEOUT)
 	fmt.Println("ONLYVERIFYPUBKEY  : ", ONLYVERIFYPUBKEY)
+
```
ameihm0912: Could we remove this extra empty line?

tydavis: Whoops. Should be fixed in 517fd01

mig-agent/agent.go
tydavis: @ameihm0912  Good point. Fixed in 517fd01

tydavis:
@ameihm0912 Fixed your commentary in 517fd01

PR #370, opened 2017-07-19
Add checkBytes function, fixes #120

### LINKED ISSUE #120: [mig modules] Support bytes string search in File module (Bugzilla #1163598)
Migrated from https://bugzilla.mozilla.org/show_bug.cgi?id=1163598

On 2015-05-11 06:57:45 -0700, Julien Vehent [:ulfr] wrote:

> The file module should allow searching for files that contain a specific hexadecimal encoded bytes string.
> 
> This feature is implemented in the memory module under the `-bytes` options, we should reuse the name in the file module.
> 
> The implementation should use a sliding window to match bytes string at the edge of search buffers, and not using line by line search as is currently the case for regexp searches.

tydavis:
Still needs go tests, but I think I can work on it through another PR

### DIFF
--- a/modules/file/file.go
+++ b/modules/file/file.go
@@ -1641,6 +1641,18 @@ func (r *run) checkBytes(f fileEntry) {
 	reader := f.getReader()
 	defer f.Close()
 
+	// Set up large buffers for reuse
+	const blocksize int = 4 * (2 << 10) // 4096, 4k
+	// Larger buffer space to allow for large block-read
+	var bigbuf = make([]byte, (blocksize + 1024))
+	var tempbuf = make([]byte, blocksize)
+
+	// Read slightly more than 4k so we have a full buffer to scan through
+	// as we move into the for loop
+	_, err = reader.Read(bigbuf)
+	if err != nil && err != io.EOF {
+		panic(err)
+	}
 	// Must use the "label,search" for loop iteration because of
 	// map[string]search for parameters
 	for _, search := range r.Parameters.Searches {
@@ -1649,15 +1661,6 @@ func (r *run) checkBytes(f fileEntry) {
 		}
 
 		for _, c := range search.checks {
-			const blocksize int = 4 * (2 << 10) // 4096, 4k
-			var bigbuf = make([]byte, (blocksize + (2 * c.bytesLength)))
-			// Read slightly more than 4k so we have a full buffer to scan through
-			// as we move into the for loop
-			_, err = reader.Read(bigbuf)
-			if err != nil && err != io.EOF {
-				panic(err)
-			}
-
 			for err != io.EOF {
 				if bytes.Contains(bigbuf, c.bytes) {
 					// We do find the intended hex values in our scan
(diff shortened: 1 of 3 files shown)


modules/file/file.go
```diff
@@ -1478,6 +1516,72 @@ func (r *run) checkHash(f fileEntry, hashtype checkType) {
...
+
+	// Must use the "label,search" for loop iteration because of
+	// map[string]search for parameters
+	for _, search := range r.Parameters.Searches {
+		if !search.isactive {
+			continue
+		}
+
+		for _, c := range search.checks {
+
```
ameihm0912: Extra newline here

modules/file/file.go
```diff
@@ -1478,6 +1516,72 @@ func (r *run) checkHash(f fileEntry, hashtype checkType) {
...
+			for err != io.EOF {
+				if bytes.Contains(bigbuf, decodedByte) {
+					// We do find the intended hex values in our scan
+					// Break the loop and return
+					// Not certain exactly how to do that here
+					c.storeMatch(f.filename)
+					break
+				} else { //Explicitly, we do not find the hex value, and have not encountered EOF
+					tempBuf := make([]byte, blocksize)
+					_, err = reader.Read(tempBuf)
```
ameihm0912: Off hand, I'm not sure if this part is correct. As an example, in this loop we will keep reading the file until we encounter EOF. The problem is we are within the loop iterating over the checks already.

So, say for example we have two bytes checks in the same action. The second will never be used to match anything since we will have read the entire file while scanning for the first check.

In the standard regex content checking function, the file reads occur outside the checks loop and each check is then applied to the read in buffer.

e.g., https://github.com/mozilla/mig/blob/35ea4f9f2f2a9bb42d7df83b9afdf55d4b544b75/modules/file/file.go#L1263-L1281

modules/file/file.go
```diff
@@ -1478,6 +1516,72 @@ func (r *run) checkHash(f fileEntry, hashtype checkType) {
...
+			_, err = reader.Read(bigbuf)
+			if err != nil && err != io.EOF {
+				panic(err)
+			}
+
+			for err != io.EOF {
+				if bytes.Contains(bigbuf, decodedByte) {
+					// We do find the intended hex values in our scan
+					// Break the loop and return
+					// Not certain exactly how to do that here
```
ameihm0912: I think what we would want to do is mark the check status as true if it has matched (e.g., checksstatus[label][i] = true and continue reading the file, so the other checks that may need to run can execute?

modules/file/file.go
```diff
@@ -1478,6 +1516,72 @@ func (r *run) checkHash(f fileEntry, hashtype checkType) {
...
+						panic(err)
+					}
+					// Take a small slice from bigbuf for sliding window
+					smallBuf := make([]byte, 2*scansize)
+					// Get the last two scan-lengths so as to permit the sliding window through End-of-Read
+					copy(smallBuf, bigbuf[blocksize:])
+					// Now make bigbuf out of two scan-lengths plus the 4k buffer read
+					var bigbuf = make([]byte, (2 * scansize))
+					copy(bigbuf, smallBuf)
+					bigbuf = append(bigbuf, tempBuf...)
```
jvehent: I'm not fully convinced this sliding window algorithm works as intended. It would be nice to summarize the logic in a comment at the top of the function, and have tests that focus on verifying it is implemented correctly.
`copy` and `make` are also expensive functions, and it would be good to call them as rarely as possible. My suspicion is this algorithm can be fine tuned further to reduce the cost of the check.

tydavis:
Doh. My original PR (before I recopied my changes into a fresh branch -- rebase failed me) had changes to `paramscreator.go` :(

modules/file/file.go
```diff
@@ -1478,6 +1516,72 @@ func (r *run) checkHash(f fileEntry, hashtype checkType) {
...
+	// Must use the "label,search" for loop iteration because of
+	// map[string]search for parameters
+	for _, search := range r.Parameters.Searches {
+		if !search.isactive {
+			continue
+		}
+
+		for _, c := range search.checks {
+
+			decodedByte := make([]byte, hex.DecodedLen(len(c.bytes)))
```
tydavis: I completely forgot about the check that provides the decode.

modules/file/file.go
```diff
@@ -1478,6 +1516,72 @@ func (r *run) checkHash(f fileEntry, hashtype checkType) {
...
+			_, err := hex.Decode(decodedByte, c.bytes)
+			if err != nil {
+				e := fmt.Sprintf("unable to decode byte query: %v ", err)
+				panic(e)
+			}
+			scansize := len(decodedByte) // Get the actual byte slice length
```
tydavis: @jvehent How would I also store the length when I decode the byte string?  Wouldn't it make more sense to determine the length of a specific check once, when it's being called?

jvehent: This is the length of the hexadecimal byte string being searched, which remains the same throughout this entire search. You should decode the hex string, calculate its length and store it once when going through the `makeChecks` function, but simply adding `bytesStr` and `length` to the `check` struct.

tydavis:
@ameihm0912 @jvehent  Still working on tests, but restructured the search to have reusable buffers (avoiding the `make` and `copy` operations).

ameihm0912:
@tydavis just wanted to check in on this one, is this still a WIP ?

tydavis:
Sorry @ameihm0912 -- life is getting in the way of my finishing this. Apologies for keeping the PR open.

tristanweir:
Closing in preparation of archiving this repository

PR #371, opened 2017-07-21
abstract Client http.Client with an interface

ameihm0912:
Introduces a new Client interface type which is used to interact with
the MIG API. The main intent of this interface is to allow substitution
of the Client type with other implementations (e.g., a Client type
suitable for unit testing).

The original Client becomes MIGClient, and satisfies the Client
interface.

A test file has been included that adds a new TestClient type which also
satisfies Client, and can be extended for internal tests.

Also, the existing http.Client member of the MIG client type has been
replaced with a requestor interface. This is intended to permit the
default client http.Client and transport configuration to be replaced if
custom options are desired.

### DIFF
--- a/client/client.go
+++ b/client/client.go
@@ -34,10 +34,88 @@ import (
 
 var version string
 
-// A Client provides all the needed functionalities to interact with the MIG API.
-// It should be initialized with a proper configuration file.
-type Client struct {
-	API     *http.Client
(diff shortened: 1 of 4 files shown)


ameihm0912:
Disregard actually, I'll resubmit here I'd like to change a couple things around

PR #378, opened 2017-08-17
file: fix bug in maxdepth handling with multiple paths in a search

ameihm0912:
In cases where multiple paths are present in a single search (e.g.,
-path is used more than once with the mig command line) the maxdepth
option was not behaving correctly.

pathWalk would examine the paths in scope for a search to determine if
it should be considered. Upon finding one, it would not break out of the
loop, so depending on the ordering, the search may not be marked as
current even if it was under the root of a search path.

This was resulting in the current directory depth not being consistently
reduced upon returning from pathWalk, causing the current depth to
continuously increase for a given search.


ameihm0912:
@jvehent wondering if you could take a quick peek at this one

PR #391, opened 2017-09-06
Addition of base dispatch and audit modules

ameihm0912:
This PR adds new persistent module related functionality to the agent.

The audit module has been added, based on [libaudit-go](https://github.com/mozilla/libaudit-go).

The audit module integrates with the Linux kernel auditing subsystem via netlink, and can be used by the agent to collect and parse the kernel audit trail. When the persistent module is started by the agent, a rule set is applied in the kernel and the module begins collecting an audit trail, sending each event to the master agent process as an "alert" message, which is a new class of message handled differently than logs.

When the master agent process receives these alerts, it can write them to it's log or in the case of the dispatch module running, the agent will forward them to the dispatch module which is also a persistent module running as a subprocess.

The dispatch module if running receives all "alert" class messages generated by modules, and buffers/dispatches them to a configured remote endpoint such as an event collector. This module currently only supports HTTP POST of incoming alert messages, but can be extended to support other transports.


modules/audit/doc.rst
```diff
@@ -0,0 +1,49 @@
...
+
+.. code::
+
+        [audit]
+        rulespath = /etc/mig/audit.rules.json
+        ratelimit = 500
+        backloglimit = 16384
+        includeraw = no
+
+``rulespath`` indicates the path to load audit rules into the kernel from. Note that this is not a
```
gdestuynder: probably add a blurb about all options, potentially by linking to audit C doc (f.e. https://people.redhat.com/sgrubb/audit/)

modules/dispatch/dispatch.go
```diff
@@ -0,0 +1,266 @@
...
+		handlerErrChan <- err
+		return
+	}
+}
+
+func dispatchIn(msg string) {
+	select {
+	case messageBuf <- msg:
+	default:
+		// If we can't queue the message it is just dropped
```
gdestuynder: shouldnt that be logged?

modules/dispatch/dispatch.go
```diff
@@ -0,0 +1,266 @@
...
+	}
+}
+
+func runDispatch(cfg config) error {
+	var (
+		httpClient http.Client
```
gdestuynder: Note, in audisp-json, I found that HTTP POSTs even when over a keep-alive connection cause buffer issues at times, as the server on the other hand sometimes cannot process that many messages that quickly, which cause the client to buffer up. Had less issues with syslog and even less with things like msgpack.

modules/dispatch/dispatch.go
ameihm0912: Agreed, I'll add some sort of feedback when this happens

PR #413, opened 2017-11-28
Delete deprecated mailing list from README

netantho:
http://list.mig.ninja/listinfo/list
Service Unavailable


jvehent:
yeah, I think that's properly dead :)

PR #432, opened 2018-02-16
Add contribute.json

claudijd:
Let's go for an A+ on the observatory, why not, right?

### DIFF
--- a/contribute.json
+++ b/contribute.json
@@ -1,0 +1,34 @@
+{
+    "name": "MIG: Mozilla InvestiGator",
+    "description": "Mozilla's real-time digital forensics and investigation platform.",
+    "repository": {
+        "url": "https://github.com/mozilla/mig",
+        "license": "MPL2",
+        "tests": "https://travis-ci.org/mozilla/mig"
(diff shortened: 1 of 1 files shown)


contribute.json
```diff
@@ -0,0 +1,34 @@
...
+    "description": "Mozilla's real-time digital forensics and investigation platform.",
+    "repository": {
+        "url": "https://github.com/mozilla/mig",
+        "license": "MPL2",
+        "tests": "https://travis-ci.org/mozilla/mig"
+    },
+    "participate": {
+        "home": "https://mig.mozilla.org/",
+        "docs": "https://mig.mozilla.org/doc.html",
+        "irc": "irc://irc.mozilla.org/#infosec",
```
ameihm0912: MIG has an IRC channel as well (#mig) but this is probably OK too

claudijd: I'm fine with fixing, didn't know.

claudijd: Fixed in cb77c6121b373a492fc59f81132fd6f0f344979d

Issue #437, opened 2018-02-27
make test failing


zsck:
Following the instructions in the [configuration.rst document](https://github.com/mozilla/mig/blob/master/doc/configuration.rst#1prepare-a-build-environment), I've tried running `make test` in the `mig/` directory. This operation failed with the following error:

```
GOOS=darwin GOARCH=amd64 GO15VENDOREXPERIMENT=1 go test mig.ninja/mig/modules/
ok  	mig.ninja/mig/modules	0.111s
GOOS=darwin GOARCH=amd64 GO15VENDOREXPERIMENT=1 go test mig.ninja/mig/modules/agentdestroy
# mig.ninja/mig/modules/agentdestroy
modules/agentdestroy/agentdestroy.go:59: Errorf format %s has arg r.Parameters.PID of wrong type int
FAIL	mig.ninja/mig/modules/agentdestroy [build failed]
make: *** [test-modules] Error 2
```

Specifically, we can see there's a simple format string error, where a string is expected but an int- `r.Parameters.PID` is provided.

zsck:
This issue has been resolved in my `issue437` branch, which I've successfully merged into `master`.

PR #438, opened 2018-02-27
Fixed format string errors that were blocking make test

zsck:
As I reported in #437, I was seeing some errors after running `make test` that were causing tests to fail.  These errors are extremely simple format string issues which seem to be being handled more strictly in more recent versions of Go (I'm using 1.10).

I'd like some feedback about how to proceed from where this PR would leave us.  My commit here fixes the format string errors I encountered, however I am now seeing some different errors. These ones appear to be more critical issues relevant to the actual function of MIG.  I would hate to submit a PR that still has errors, but hope we can discuss how to proceed. It may be desirable to use the branch I've created as a pivot around which we'll commit more fixes before merging into `master`.

```
OOS=darwin GOARCH=amd64 GO15VENDOREXPERIMENT=1 go test mig.ninja/mig
--- FAIL: TestVerifyACLValid (0.00s)
	action_test.go:225: VerifySignatures: VerifySignatures() -> openpgp: invalid signature: hash tag doesn't match
--- FAIL: TestVerifyGoodSigBadACL (0.00s)
	action_test.go:297: VerifySignatures: VerifySignatures() -> openpgp: invalid signature: hash tag doesn't match
--- FAIL: TestVerifyBadSig (0.00s)
	action_test.go:394: VerifySignatures: VerifySignatures() -> openpgp: invalid signature: hash tag doesn't match
--- FAIL: TestVerifyMultiOperation (0.00s)
	action_test.go:440: VerifySignatures: VerifySignatures() -> openpgp: invalid signature: hash tag doesn't match
FAIL
FAIL	mig.ninja/mig	0.016s
```

### DIFF
--- a/client/mig-console/action_launcher.go
+++ b/client/mig-console/action_launcher.go
@@ -357,8 +357,7 @@ times			show the various timestamps of the action
 				fmt.Println(`Invalid times. Expects settimes <start> <stop.)
 examples:
 settimes 2014-06-30T12:00:00.0Z 2014-06-30T14:00:00.0Z
-settimes now +60m
-`)
+settimes now +60m`)
 				break
 			}
 			if orders[1] == "now" {
@@ -366,7 +365,7 @@ settimes now +60m
 				a.ValidFrom = time.Now().Add(-60 * time.Second).UTC()
 				period, err := time.ParseDuration(orders[2])
 				if err != nil {
-					fmt.Println("Failed to parse duration '%s': %v", orders[2], err)
+					fmt.Printf("Failed to parse duration '%s': %v\n", orders[2], err)
 					break
 				}
 				a.ExpireAfter = a.ValidFrom.Add(period)
@@ -374,12 +373,12 @@ settimes now +60m
 			} else {
 				a.ValidFrom, err = time.Parse("2014-01-01T00:00:00.0Z", orders[1])
 				if err != nil {
-					fmt.Println("Failed to parse time '%s': %v", orders[1], err)
+					fmt.Printf("Failed to parse time '%s': %v\n", orders[1], err)
 					break
 				}
 				a.ExpireAfter, err = time.Parse("2014-01-01T00:00:00.0Z", orders[2])
 				if err != nil {
-					fmt.Println("Failed to parse time '%s': %v", orders[2], err)
+					fmt.Printf("Failed to parse time '%s': %v\n", orders[2], err)
 					break
 				}
 			}
--- a/client/mig-console/investigator.go
+++ b/client/mig-console/investigator.go
@@ -320,10 +320,10 @@ func investigatorCreator(cli client.Client) (err error) {
 	}
 	switch strings.ToLower(respv) {
 	case "yes":
-		fmt.Println("Investigator will have manifest management permissions\n")
+		fmt.Println("Investigator will have manifest management permissions")
 		pset.ManifestSet()
 	case "no":
-		fmt.Println("Investigator will not have manifest management permissions\n")
+		fmt.Println("Investigator will not have manifest management permissions")
 	default:
 		panic("must specify yes or no")
 	}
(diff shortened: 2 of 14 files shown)


client/mig-console/action_launcher.go
```diff
@@ -357,29 +357,28 @@ times			show the various timestamps of the action
 				fmt.Println(`Invalid times. Expects settimes <start> <stop.)
 examples:
 settimes 2014-06-30T12:00:00.0Z 2014-06-30T14:00:00.0Z
-settimes now +60m
-`)
```
zsck: I was seeing an error (not a warning, surprisingly) here due to the "redundant newline" at the end of this format string.

mig-api/action_endpoints.go
```diff
@@ -46,7 +46,7 @@ func createAction(respWriter http.ResponseWriter, request *http.Request) {
 	if err != nil {
 		panic(err)
 	}
-	ctx.Channels.Log <- mig.Log{OpID: opid, Desc: fmt.Sprintf("Received action for creation '%s'", action)}.Debug()
+	ctx.Channels.Log <- mig.Log{OpID: opid, Desc: fmt.Sprintf("Received action for creation '%v'", action)}.Debug()
```
zsck: This is a change I'd like to be particularly mindful of.  It looks like we're "just" logging here, so I didn't worry much about this change. The `%v` format here results in the traditional behaviour such that:

```go
package main

import (
	"fmt"
)

type Thing struct {
	Id int
	Name string
}

func main() {
	fmt.Printf("Thing %v", Thing { 0, "name" })
}
```

produces:

> Thing {0 name}

Whereas replacing `%v` with `%s` produces:

> Thing {%!s(int=0) name}

mig-scheduler/pgp.go
```diff
@@ -141,6 +141,6 @@ func makeSchedulerInvestigator(orig_ctx Context) (inv mig.Investigator, err erro
 	if err != nil {
 		panic(err)
 	}
-	ctx.Channels.Log <- mig.Log{Desc: fmt.Sprintf("created migscheduler identity with ID %d and key ID %s", iid, inv.PGPFingerprint)}
```
zsck: `iid` is a `float64`

modules/agentdestroy/agentdestroy.go
```diff
@@ -56,7 +56,7 @@ type results struct {
 
 func (r *run) ValidateParameters() (err error) {
 	if r.Parameters.PID < 2 || r.Parameters.PID > 65535 {
-		return fmt.Errorf("PID '%s' is not in the range [2:65535]", r.Parameters.PID)
```
zsck: `PID` is an `int`.

client/mig-console/investigator.go
```diff
@@ -320,10 +320,10 @@ func investigatorCreator(cli client.Client) (err error) {
 	}
 	switch strings.ToLower(respv) {
 	case "yes":
-		fmt.Println("Investigator will have manifest management permissions\n")
+		fmt.Println("Investigator will have manifest management permissions")
```
ameihm0912: I'm not sure off hand but this one might have been intentional to break up the output into logical blocks during investigator creation

client/mig-console/investigator.go
```diff
@@ -320,10 +320,10 @@ func investigatorCreator(cli client.Client) (err error) {
...
 		pset.ManifestSet()
 	case "no":
-		fmt.Println("Investigator will not have manifest management permissions\n")
+		fmt.Println("Investigator will not have manifest management permissions")
```
ameihm0912: Same here

zsck:
This latest commit I've pushed looks to both

* Make the compiler happy with regards to the "redundant newlines" it was complaining to me about and
* Give us clean spacing in the tool's output

PR #455, opened 2018-07-06
Check /etc/os-release for the host's ident if /etc/issue contains only \S

zsck:


mig-agent/agentcontext/env_linux.go
```diff
@@ -75,11 +75,21 @@ func findOSInfo(orig_ctx AgentContext) (ctx AgentContext, err error) {
...
 	}
 	logChan <- mig.Log{Desc: fmt.Sprintf("getIssue() failed: %v", err)}.Debug()
+
+	ctx.OSIdent, err = getOSRelease()
+	if err == nil {
+		logChan <- mig.Log{Desc: "using /etc/os-release for distribution ident"}.Debug()
+		goto haveident
+	}
+
 	logChan <- mig.Log{Desc: "warning, no valid linux os identification could be found"}.Info()
```
ameihm0912: Should this just return at this point?

mig-agent/agentcontext/env_linux.go
```diff
@@ -75,11 +75,21 @@ func findOSInfo(orig_ctx AgentContext) (ctx AgentContext, err error) {
...
+
+	// Here we check that we read more than '\S'.
+	// See https://access.redhat.com/solutions/1138953
+	if err == nil && len(ctx.OSIdent) > 3 {
 		logChan <- mig.Log{Desc: "using /etc/issue for distribution ident"}.Debug()
 		goto haveident
 	}
 	logChan <- mig.Log{Desc: fmt.Sprintf("getIssue() failed: %v", err)}.Debug()
+
+	ctx.OSIdent, err = getOSRelease()
```
ameihm0912: It doesn't look like we log the error response here like we do with getIssue()

zsck: Right you are! This is fixed.

mig-agent/agentcontext/env_linux.go
zsck: This function also populates the `AgentContext`'s `Init` field a couple lines down from here.

zsck:
@ameihm0912 Going to request a re-review when you have time.  Any idea what could be causing the builds in Travis to fail?``

zsck:
Errors seem to be changing all the time.

I'm not sure why there's a problem importing the `osext` module.  I've been able to run all the mig tests locally just fine with

```
go clean --testcache
make test
```

PR #460, opened 2018-08-13
[WIP] Propose install script to setup mig

claudijd:
I was asked to install mig for dog-fooding purposes and found the manual commands to be annoying and would like to propose something like this to simplify it for the average user who's being asked to install mig so it's not as much work and just speeds up adoption.

I didn't add any code, because the code required is trivial, this is more about proposing the concept.  If you'd prefer I actually write the code and would want to port installation docs to reflect this install stuff, then I'd be willing to that big for you.


tristanweir:
Closing in preparation of archiving this repository

PR #463, opened 2018-08-22
Establish a build pipeline to push raw binaries to S3

zsck:
This PR includes a few new things to facilitate a new build pipeline.  The goal is to start having TravisCI start produce builds of the MIG Agent and to have those builds uploaded to an S3 bucket for anyone to download. This should help to facilitate a move away from the complicated build and release process we currently follow to produce unique builds for each team at Mozilla.

Changes include:

1. A new mig_agent_cfg tool that produces a `mig-agent.cfg` file's contents given some command-line flags
2. A new build-agent-release.sh tool that will cross-compile the agent and the mig_agent_cfg tool for each platform we want to be able to deploy to (all six combinations of Linux, Darwin and Windows builds for i386 and x86_64)
3. A new releases.json file within which we will record information about builds of the agent and agent config tool so that consumers can quickly determine the name of the latest release and get a sha256 hash of the binary
4. A new update_release_json tool that will update releases.json given information about a specific release
5. An update to our TravisCI configuration to have it deploy everything built by the above three tools as well as the releases.json file to a public S3 bucket

Note that Travis appears to not push to S3 from branches other than `master`. So, once this change lands, we'll have to verify that that works.


ameihm0912:
@zsck just a couple comments here for consideration

There are a few reasons we have avoided cross-compilation in the past:

* Some modules use cgo, so cross-compiling here will not work for them. It looks like there might be a few ways to work around this by installing platform specific build tool chains, but have never gone down this path.
* There are platform specific tests, but this will just run the tests on Linux which may not be the route we'd want to go if the Darwin/Windows builds are going to be deployed.
* Darwin requires a signature on the binary via codesign to properly execute.

Makefile
```diff
@@ -141,7 +142,7 @@ mig-runner: create-bindir
 	$(GO) build $(GOOPTS) -o $(BINDIR)/mig-runner $(GOLDFLAGS) github.com/mozilla/mig/mig-runner
 
 mig-action-generator: create-bindir
-	$(GO) build $(GOOPTS) -o $(BINDIR)/mig-action-generator $(GOLDFLAGS) github.com/mozilla/mig/client/mig-action-generator
+	S(GO) build $(GOOPTS) -o $(BINDIR)/mig-action-generator $(GOLDFLAGS) github.com/mozilla/mig/client/mig-action-generator
```
floatingatoll: `S(GO)` may be a typo.

Makefile
```diff
@@ -124,6 +124,7 @@ create-bindir:
 mig-agent: create-bindir
 	@echo building mig-agent for $(OS)/$(ARCH)
 	$(GO) build $(GOOPTS) -o $(BINDIR)/mig-agent-$(BUILDREV)$(BINSUFFIX) $(GOLDFLAGS) github.com/mozilla/mig/mig-agent
+	$(GO) build $(GOOPTS) -o $(BINDIR)/mig_agent_cfg tools/mig_agent_cfg.go
```
floatingatoll: This `mig_agent_cfg` binary uses underscores when all others seem to use hyphens. If it’s introduced newly in this patch, consider altering it to match if that’s appropriate.

releases/releases.json
```diff
@@ -0,0 +1,11 @@
+{
+    "agent": {
+        "latest": "",
+        "releases": [
+        ]
+    },
+    "agentConfig": {
+        "latest": "",
+        "releases": []
```
floatingatoll: Should line 4-5 and line 9 use the same whitespace approach for the empty list?

zsck:
@ameihm0912 If I were to restrict this to just building for Linux, how would things play out?

tools/build-agent-release.sh
```diff
@@ -0,0 +1,54 @@
+RELEASEDIR=releases
+TAG=$(date +%Y%m%d)-0.$(git log --pretty=format:'%h' -n 1)
```
floatingatoll: Consider some error checking of these two subcommands’ output, since this is a convenient time (at line 2) to detect if ‘git log’ isn’t working properly. (Unlikely, but if you do end up with a tag ending in “0.” someday, this is why!)

tools/build-agent-release.sh
```diff
@@ -0,0 +1,54 @@
...
+
+BUILDS=(
+  # FORMAT
+  # GOOS GOARCH BINSUFFIX
+  "darwin 386 darwin-i386-$TAG.prod"
```
floatingatoll: What meaning is intended by “prod”? Typically I see it paired with “dev” and/or “stage”, in which case this would benefit from being moved to a variable at the top of the script so that it’s not a buried string constant.

tools/build-agent-release.sh
```diff
@@ -0,0 +1,54 @@
...
+# 4. The path to the new agent binary.
+function update_releases() {
+  releasejson=$(go run tools/update_release_json.go \
+    -releases $RELEASEDIR/$1 \
+    -component $2 \
+    -tag $3 \
+    -binary $4 \
+    -sha256 $(cat $RELEASEDIR/$4 | openssl sha256))
+
+  echo "$releasejson" > $RELEASEDIR/$1
```
floatingatoll: It would be nice to verify that the release JSON is valid JSON here, in case one of the subcommands above breaks for some reason.

tools/build-agent-release.sh
```diff
@@ -0,0 +1,54 @@
...
+
+# update_releases invokes the update_release_json tool to record an agent update.
+# It expects to be called with the following arguments:
+# 1. The path to the releases.json file to update,
+# 2. The component to update,
+# 3. The tag for the new release and
+# 4. The path to the new agent binary.
+function update_releases() {
+  releasejson=$(go run tools/update_release_json.go \
+    -releases $RELEASEDIR/$1 \
```
floatingatoll: These bare variables are not whitespace-safe, which is fine if none of them will ever contain spaces. Consider running `shellcheck` against this script.

tools/build-agent-release.sh
```diff
@@ -0,0 +1,54 @@
+RELEASEDIR=releases
```
floatingatoll: Please add a `#!` line for bash, since you use bash-specific features below.

tools/build-agent-release.sh
```diff
@@ -0,0 +1,54 @@
...
+
+
+for build in "${BUILDS[@]}"; do
+  IFS=" " read -ra args <<< "$build"
+
+  echo "Building an agent for ${args[0]}/${args[1]}"
+  GOOS=${args[0]} GOARCH=${args[1]} go build -o $RELEASEDIR/mig-agent-${args[2]} github.com/mozilla/mig/mig-agent
+  if [ $? -eq 0 ]; then
+    echo "+ Build succeeded"
+    update_releases releases.json agent $TAG mig-agent-${args[2]}
```
floatingatoll: Is any other file than `releases.json` ever updates with this function? If not, this might be better off as a top-level variable instead of a buried string constant.

tools/build-agent-release.sh
```diff
@@ -0,0 +1,54 @@
...
+    -sha256 $(cat $RELEASEDIR/$4 | openssl sha256))
+
+  echo "$releasejson" > $RELEASEDIR/$1
+}
+
+
+for build in "${BUILDS[@]}"; do
+  IFS=" " read -ra args <<< "$build"
+
+  echo "Building an agent for ${args[0]}/${args[1]}"
```
floatingatoll: I would encourage setting `targetarch=` here first, and then including it in both in this debug message and also the “failed” message below so that a simple build log grep for “failed” shows precisely what failed without having to dig for more context.

ameihm0912:
I think doing the Linux build here would be fine

No reason we can't leave the other builds in here too if that's the route you want to go, just wanted to point those small gotchas out. There will be feature discrepancies in the resulting binaries, so we've avoided this because it isn't a complete solution.

The ideal scenario would be doing builds/tests on the native platforms themselves, but this requires more machinery to do.

tools/mig_agent_cfg.go
```diff
@@ -0,0 +1,274 @@
...
+    installservice   = {{.InstallAsService}}
+
+    ; attempt to retrieve the public IP behind which the agent is running
+    discoverpublicip = on
+
+    ; in check-in mode, the agent connects to the relay, runs all pending commands
+    ; and exits. this mode is used to run the agent as a cron job, not a daemon.
+    checkin = {{.CronMode}}
+
+    refreshenv = "5m"
```
floatingatoll: This entry is missing documentation in the template.

tools/mig_agent_cfg.go
```diff
@@ -0,0 +1,274 @@
...
+type tagList []string
+
+// Config contains user-supplied configurable values that will be translated
+// and injected into `release/mig-agent.cfg.template` to produce a valid
+// `mig-agent.cfg`.
+type Config struct {
+	RelayAddress     string
+	APIURL           string
+	Proxies          proxyList
+  Tags             tagList
```
floatingatoll: This line has whitespace issues versus its surrounding lines.

tools/mig_agent_cfg.go
```diff
@@ -0,0 +1,274 @@
...
+}
+
+// translatedConfig contains the same data as `Config` with some
+// transformations to make the output of this program consistent
+// with what mig agents expect.
+type translatedConfig struct {
+	RelayAddress     string
+	APIURL           string
+	Proxies          string
+  Tags             map[string]string
```
floatingatoll: This line has whitespace issues versus its surrounding lines.

tools/mig_agent_cfg.go
```diff
@@ -0,0 +1,274 @@
...
+	ExtraPrivacy     string
+}
+
+// onOrOff converts a bool from `true` to `"on"` and `false` to `"off"`.
+func onOrOff(on bool) string {
+	if on {
+		return "on"
+	} else {
+		return "off"
+	}
```
floatingatoll: The preceding lines of this script have different whitespace than the following lines - 8-space versus 2-space. Consider normalizing to one form only.

tools/mig_agent_cfg.go
```diff
@@ -0,0 +1,274 @@
...
+  tags, err := kvPairsToMap(config.Tags)
+  if err != nil {
+    return translatedConfig{}, err
+  }
+
+  tx := translatedConfig{
+		RelayAddress:     config.RelayAddress,
+		APIURL:           config.APIURL,
+		Proxies:          strings.Join([]string(config.Proxies), ","),
+    Tags:             tags,
```
floatingatoll: This line has whitespace issues versus its surrounding lines.

tools/mig_agent_cfg.go
```diff
@@ -0,0 +1,274 @@
...
+// Set is required by the `flag.Value` interface. We use it to enable users to
+// supply multiple proxy addresses, which we collect into an array.
+func (proxies *proxyList) Set(value string) error {
+	*proxies = append(*proxies, value)
+	return nil
+}
+
+// String is required by the `flag.Value` interface.
+func (tags *tagList) String() string {
+  return strings.Join([]string(*tags), ",")
```
floatingatoll: The code resumes 2-space indenting from 8-space indenting here, unexpectedly.

tools/mig_agent_cfg.go
```diff
@@ -0,0 +1,274 @@
...
+		"Instruct the agent to run with extra privacy controls")
+	flag.Parse()
+
+	template := template.Must(template.New("mig-agent.cfg").Parse(configTemplate))
+
+	config := Config{
+		RelayAddress:     *relayAddr,
+		APIURL:           *apiUrl,
+		Proxies:          proxies,
+    Tags:             tags,
```
floatingatoll: This line has whitespace issues versus its surrounding lines.

tools/update_release_json.go
```diff
@@ -0,0 +1,171 @@
...
+
+	if *tag == "" {
+		fmt.Fprintf(os.Stderr, "No tag identifier supplied.\n")
+		os.Exit(1)
+	}
+
+	setLatestRelease(component, *binary, *tag, *notes, *sha256)
+
+	prefix := ""
+	indent := "    "
```
floatingatoll: Ironically, the 4-space indent used in the JSON is not used anywhere in this source code otherwise, which uses only 2- or 8-. That’s not a criticism, but it does count as a third indenting style in use in a single file, so duly noted.

tools/update_release_json.go
```diff
@@ -0,0 +1,171 @@
...
+// knownComponents returns a list of names of components understood by this
+// tool, for which new releases can be documented.
+func knownComponents() []string {
+	return []string{
+		"agent",
+		"agentConfig",
+	}
+}
+
+// component retrievses a pointer to a known component identified by its name,
```
floatingatoll: Typo of “retrieves”.

floatingatoll:
The Taskcluster project within Mozilla offers cross-platform build and CI
with GitHub auto-integration for specifically the platforms you’re having
issues targeting, if/when that becomes a serious obstacle to progress.

tools/build-agent-release.sh
```diff
@@ -0,0 +1,54 @@
...
+  GOOS=${args[0]} GOARCH=${args[1]} go build -o $RELEASEDIR/mig-agent-${args[2]} github.com/mozilla/mig/mig-agent
+  if [ $? -eq 0 ]; then
+    echo "+ Build succeeded"
+    update_releases releases.json agent $TAG mig-agent-${args[2]}
+  else
+    echo "- Build failed";
```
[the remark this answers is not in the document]
tristanweir: Maybe it is already taken care of by Travis. My thought is it would be good to provide debug on why the build failed, not just "Build failed". 

I'm not familiar enough with how Travis would pop that info, so if it is doing something instructive enough that we can go fix whatever broke, then that's fine.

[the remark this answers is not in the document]
zsck: I believe Travis would log the failures to stderr, since we don't redirect the output of the `go build` command here.

PR #468, opened 2018-09-18
Add check for ping during postgresql connection

pwnbus:
This improves troubleshooting, or else the connection seems to be established, but realistically, no connection attempt was made, the open() function only checks syntax of the db string.

### DIFF
--- a/mig-api/context.go
+++ b/mig-api/context.go
@@ -11,6 +11,7 @@ import (
 	"io"
 	"github.com/mozilla/mig"
 	migdb "github.com/mozilla/mig/database"
+	"os"
 	"strconv"
 	"strings"
 	"sync"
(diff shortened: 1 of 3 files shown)


mig-api/context.go
```diff
@@ -113,6 +114,7 @@ func initDB(orig_ctx Context) (ctx Context, err error) {
 		ctx.Channels.Log <- mig.Log{Desc: "leaving initDB()"}.Debug()
 	}()
 
+	fmt.Fprintf(os.Stdout, "Attempting to connect to postgresql database...")
```
pwnbus: I thought it would be useful to add some sort of debug statement to print stating a connection is attempting to be made. Unfortunately the Ping function will take ~30 seconds to timeout, so the mig-api/scheduler could look like it's hanging.

mig-scheduler/context.go
```diff
@@ -195,6 +196,7 @@ func initDB(orig_ctx Context) (ctx Context, err error) {
 		ctx.Channels.Log <- mig.Log{Desc: "leaving initDB()"}.Debug()
 	}()
 
+	fmt.Fprintf(os.Stdout, "Attempting to connect to postgresql database...")
```
pwnbus: Same comment as above, some sort of feedback to the user?

mig-api/context.go
pwnbus: I believe at this point, we're unable to use "ctx.Channels.Log <- mig.Log"  to log, since the logger isn't setup yet?

pwnbus: I tried using:
```
ctx.Channels.Log <- mig.Log{Desc: "Attempting to connect to db"}"
```
however, it didn't show up in the output?

jvehent:
I'd recommend doing a 'SELECT 1' instead of using Ping, it's a more
accurate test as it actually verifies that you're able to run queries.

pwnbus:
@jvehent Updated to use SELECT 1 instead of Ping.

PR #471, opened 2018-10-19
Adding support for proxy disabling in gozdef

zsck:
~This PR is dependent on the [oustanding PR for gozdef](https://github.com/jvehent/gozdef/pull/6) to add support to disable the use of a proxy in Gozdef.~

Julien has stated that he no longer intends to maintain gozdef, so I guess I'm inheriting it.  I've forked it to `mozilla/gozdef` and pushed my changes right into master.

### DIFF
--- a/runner-plugins/runner-scribe/main.go
+++ b/runner-plugins/runner-scribe/main.go
@@ -17,9 +17,9 @@ import (
 	"strings"
 
 	"github.com/jvehent/gozdef"
-	"gopkg.in/gcfg.v1"
 	"github.com/mozilla/mig"
 	scribemod "github.com/mozilla/mig/modules/scribe"
+	"gopkg.in/gcfg.v1"
 )
 
 // config represents the configuration used by runner-scribe, and is read in on
@@ -28,8 +28,9 @@ import (
 // URL and Source are mandatory settings
 type config struct {
 	MozDef struct {
-		URL    string // URL to post events to MozDef
-		Source string // Source identifier for vulnerability events
+		URL      string // URL to post events to MozDef
+		Source   string // Source identifier for vulnerability events
+		UseProxy bool
 	}
 }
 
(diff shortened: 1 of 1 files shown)


runner-plugins/runner-scribe/main.go
```diff
@@ -17,9 +17,9 @@ import (
 	"strings"
 
 	"github.com/jvehent/gozdef"
-	"gopkg.in/gcfg.v1"
 	"github.com/mozilla/mig"
 	scribemod "github.com/mozilla/mig/modules/scribe"
+	"gopkg.in/gcfg.v1"
```
floatingatoll: Is this ordering change a necessary component of this commit, or merely aesthetic?

runner-plugins/runner-scribe/main.go
```diff
@@ -85,7 +86,10 @@ func main() {
 }
 
 func sendVulnerability(item gozdef.VulnEvent) (err error) {
-	ac := gozdef.APIConf{URL: conf.MozDef.URL}
+	ac := gozdef.APIConf{
```
floatingatoll: This should be `APIConf {` per surrounding examples (such as line 110).

runner-plugins/runner-scribe/main.go
```diff
@@ -28,8 +28,9 @@ import (
 // URL and Source are mandatory settings
 type config struct {
 	MozDef struct {
-		URL    string // URL to post events to MozDef
-		Source string // Source identifier for vulnerability events
+		URL      string // URL to post events to MozDef
+		Source   string // Source identifier for vulnerability events
+		UseProxy bool
```
floatingatoll: Consider adding a `// comment` line or removing the two above.

runner-plugins/runner-scribe/main.go
zsck: This reorder was done automatically by `gofmt`, which prefers imports be listed alphabetically.

runner-plugins/runner-scribe/main.go
zsck: This is the format enforced by `gofmt`.

runner-plugins/runner-scribe/main.go
zsck: Fair suggestion! Done.

floatingatoll:
That’s weird - I would have expected it to apply the same elsewhere. Oh
well!

zsck:
I suspect that what's happening is my `vim-go` plugin is only formatting the code I changed, which in this case is kind of nice.  I could rip through everything and format all our code, but that's not really my goal here.

floatingatoll:
No worries!

PR #472, opened 2018-10-29
Runner scribe config

zsck:
This branch contains some unit tests I wrote to verify that the changes I made to the `runner-scribe` plugin were correct.


modules/modules_test.go
```diff
@@ -183,9 +183,8 @@ func TestReadInputParameters(t *testing.T) {
 func TestWatchForStop(t *testing.T) {
 	stopChan := make(chan bool)
 	w := NewModuleReader(strings.NewReader(`{"class":"stop"}`))
-	var err error
 	go func() {
-		err = WatchForStop(w, &stopChan)
```
zsck: Removed this because I was getting a compilation error regarding the variable remaining unused.

PR #473, opened 2018-11-09
Configure heartbeat requests to go straight to API

pwnbus:
This bypasses rabbitmq entirely for heartbeats. The agents will now post heartbeats to the API directly, which writes to postgres.


database/actions.go
```diff
@@ -158,6 +202,19 @@ func (db *DB) InsertAction(a mig.Action) (err error) {
...
 		return fmt.Errorf("Failed to store action: '%v'", err)
 	}
+
+	// The following query establishes a relation between the new action and all
+	// agents targeted by the action.
+	_, err = db.c.Exec(fmt.Sprintf(`
+  insert into agent_action_relation (agent_id, action_id)
+  select $1, A.id
+  from agents A
+  where (%s);
```
jvehent: isn't that a sql injection?

jvehent: I'm an idiot... of course it's a sql injection, that's what the target is in the first place.
Sorry about that, it's been years since I looked at that code

database/schema.sql
```diff
@@ -42,6 +42,16 @@ CREATE INDEX agents_starttime_idx ON agents(starttime DESC);
 CREATE INDEX agents_queueloc_pid_idx ON agents(queueloc, pid);
 CREATE INDEX agents_status_idx ON agents(status);
 
+CREATE TABLE agent_action_relation (
```
pwnbus: I think we aren't using this relationship anymore, right? @zsck

mig-scheduler/agents_queues.go
```diff
@@ -6,167 +6,13 @@
...
-		// agent authorization failed so we drop this heartbeat and return
-		return
-	}
-
-	// write to database in a goroutine to avoid blocking
-	go func() {
-		// if an agent already exists in database, we update it, otherwise we insert it
-		agent, err := ctx.DB.AgentByQueueAndPID(agt.QueueLoc, agt.PID)
-		if err != nil {
-			agt.DestructionTime = time.Date(9998, time.January, 11, 11, 11, 11, 11, time.UTC)
```
pwnbus: I looked into why the tests are failing, and it turns out some of the fields in the db are no longer getting added:
```
destructiontime | 0001-01-01 00:00:00+00
heartbeattime   | 0001-01-01 00:00:00+00
refreshtime     | 0001-01-01 00:00:00+00
```
I looked at what a good agent document looks like, and it's like so:
```
destructiontime | 9998-01-11 11:11:11+00
heartbeattime   | 2018-11-12 21:30:36.123557+00
refreshtime     | 2018-11-12 21:28:06.215546+00
```

I think we lost some of the logic here when moving this to the API FYI @zsck

mig-api/api.go
```diff
@@ -70,6 +71,13 @@ func main() {
 	r := mux.NewRouter()
 	s := r.PathPrefix(ctx.Server.BaseRoute).Subrouter()
 
+	postHeartbeat := agents.NewUploadHeartbeat(
+		agents.NewPersistHeartbeatPostgres(&ctx.DB),
+		agents.NewNilAuthenticator())
+
+	// Endpoints that replace previously direct-to-rabbitmq communications.
+	s.Handle("/heartbeat", postHeartbeat).Methods("POST")
```
pwnbus: Originally these heartbeat messages required auth to rabbitmq, so should we maintain that requirement (that an agent needs to auth somehow to post a heartbeat?)

mig-scheduler/agents_queues.go
pwnbus: The status field is also not set, when in the other one, it's set to 'online'. I think we're missing some function here to identify which agents are online and which are offline.

mig-api/api.go
zsck: I don't think it's really worth finding a way to authenticate agents in this case. If an agent or some actor imitating an agent pushes a heartbeat, they aren't going to be able to do anything really damaging. I think finding a way to authenticate agents for this purpose and implementing it would be far more work than it's worth.

database/schema.sql
zsck: Right you are. I've removed it.

mig-scheduler/agents_queues.go
zsck: Good callout.  I didn't think it was going to matter. In fact, I'm not sure our tests are failing because of this.  I've pushed a fix to include these fields and we still see things failing in Travis.

zsck: Looks to me like the tests might be failing due to them being run [before Postgres is fully initialized](https://travis-ci.org/mozilla/mig/builds/454728538#L2095). That's my hypothesis based on this error, anyway.

pwnbus: I forgot a key part, when I was doing my testing, I had merged https://github.com/mozilla/mig/pull/467 into my local branch, so my test failures were different than what travisci was reporting.

PR #476, opened 2018-12-17
Query ServiceAPI for Operator & Team

tristanweir:
This restores long-desired functionality to query ServiceAPI for an Operator and Team and then publish those to the vulnerability event being posted to MozDef.

Specifically we:

1. Get an authentication token from Auth0
2. Get all assets that ServiceAPI knows about and store them in a searchable map, keyed to hostname
3. As we encounter a host with vulnerabilities, we lookup operator and team and append them to the vulnerability event. If operator and team are not found in the ServiceAPI data, we set operator based on the config of the agent. Team remains blank.

### DIFF
--- a/runner-plugins/runner-scribe/main.go
+++ b/runner-plugins/runner-scribe/main.go
@@ -3,6 +3,7 @@
 // file, You can obtain one at http://mozilla.org/MPL/2.0/.
 //
 // Contributor: Aaron Meihm ameihm@mozilla.com [:alm]
+// Contributor: Tristan Weir tweir@mozilla.com [:weir]
 
 // runner-scribe is a mig-runner plugin that processes results coming from automated
 // actions and forwards the results as vulnerability events to MozDef
@@ -13,6 +14,7 @@ import (
 	"fmt"
 	"io/ioutil"
 	"net"
+	"net/http"
 	"os"
 	"strings"
 
@@ -33,6 +35,34 @@ type config struct {
 		URL      string // URL to post events to MozDef
 		UseProxy bool   // A switch to enable/disable the use of a system-configured proxy
 	}
+	api 		ServiceApi
+}
+
+type ServiceApiAsset struct {
+	Id 				string
+	Asset_type 		string
+	Asset_identifier string
+	Team 			string
+	Operator 		string
+	Zone 			string
+	Timestamp 		string
+	Description 	string
+	Score 			int
+}
+
+type ServiceApi struct {
+	URL				string
+	AuthEndpoint 	string
+	ClientID 		string
+	ClientSecret	string
+	Token 			string // ephemeral token we generate to connect to ServiceAPI
+}
+
+type Auth0Token struct {
+	Access_token	string
+	Scope			string
+	Expires_in		int
+	Token_type		string
 }
 
 const configPath string = "/etc/mig/runner-scribe.conf"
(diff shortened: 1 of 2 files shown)


runner-plugins/runner-scribe/main.go
```diff
@@ -209,6 +256,94 @@ func makeVulnerability(initems []gozdef.VulnEvent, cmd mig.Command) (items []goz
...
+	req.Header.Add("accept", "application/json")
+	req.Header.Add("Authorization", api.Token)
+	res, err := http.DefaultClient.Do(req)
+	if err != nil {
+		return err
+	}
+
+	// unpack the HTTP request response
+	defer res.Body.Close()
+	body, readErr := ioutil.ReadAll(res.Body)
```
arcrose: Rather than reading the entire response body into memory before trying to unmarshal it from JSON, we should use an [encoding/json.Decoder](https://golang.org/pkg/encoding/json/#Decoder).

runner-plugins/runner-scribe/main.go
```diff
@@ -209,6 +256,94 @@ func makeVulnerability(initems []gozdef.VulnEvent, cmd mig.Command) (items []goz
...
+	// convert json into array of ServiceApiAsset objects
+	var allAssets []ServiceApiAsset
+	err  = json.Unmarshal([]byte(allAssetsJson), &allAssets)
+	if err != nil {
+		return err
+	}
+
+	// build a searchable map, keyed on Asset_identifier (which is usually hostname)
+	for _, tempAsset := range allAssets {
+		permanentAsset := tempAsset				//not sure if this is needed
```
arcrose: This is not needed.  Go will copy strings for us. 👍

runner-plugins/runner-scribe/main.go
```diff
@@ -209,6 +256,94 @@ func makeVulnerability(initems []gozdef.VulnEvent, cmd mig.Command) (items []goz
...
+	}
+
+	return
+}
+
+// return the operator and team for a given hostname, provided they are in the map of 
+// ServiceApiAssets. If they are not in the map or if the values are not present, 
+// operator and/or team will return as an empty string ""
+func LookupOperatorTeam(hostname string, m map[string]ServiceApiAsset) (operator string, team string) {
+
```
arcrose: Nit: Remove blank line.

runner-plugins/runner-scribe/main.go
```diff
@@ -209,6 +256,94 @@ func makeVulnerability(initems []gozdef.VulnEvent, cmd mig.Command) (items []goz
...
+	return
+}
+
+// return the operator and team for a given hostname, provided they are in the map of 
+// ServiceApiAssets. If they are not in the map or if the values are not present, 
+// operator and/or team will return as an empty string ""
+func LookupOperatorTeam(hostname string, m map[string]ServiceApiAsset) (operator string, team string) {
+	
+	operator = m[hostname].Operator
```
arcrose: Is it possible that we may lookup a `hostname` that isn't in the map?  Since we have this abstracted under a function anyway, I propose we make the interface a little more friendly.  Perhaps something like the following.

```go
type Team struct {
  Name string
  Operator string
}

func LookupOperatorTeam(
  hostname string,
  mapping map[string]ServiceApiAsset,
) (Team, bool) {
  asset, found = mapping[hostname]
  if !found {
    return Team{}, false
  }
  team := Team{
    Name: asset.Team,
    Operator: asset.Operator,
  }
  return team, true
}

**Comment by arcrose** [comment]
I strongly suggest we write as many tests as possible for this.  It should be fairly straightforward to mock the services we're calling into using [net/http/httptest](https://golang.org/pkg/net/http/httptest/).

File: runner-plugins/runner-scribe/main.go (hunk 4 above)
Comment (tristanweir) [reply runner-plugins/runner-scribe/main.go]: In testing, if hostname is not found in the map, Operator and Team return as empty strings. This is expected behavior and we'll just set them as "" in the mozdef event. Is there anything useful we could do with the returned bool?

File: runner-plugins/runner-scribe/main.go [hunk 5]
```diff
@@ -33,6 +35,34 @@ type config struct {
...
+	Team 			string
+	Operator 		string
+	Zone 			string
+	Timestamp 		string
+	Description 	string
+	Score 			int
+}
+
+type ServiceApi struct {
+	URL				string
```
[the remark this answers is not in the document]
tristanweir: There are challenges with doing this for the config itself. The gcfg package that reads the config runner-scribe.conf file and populates the config object doesn't know how to handle objects like that. I'll do a manual cast when we actually modify the URL at the top of GetAssets, but I'll keep them stored as strings for now.

runner-plugins/runner-scribe/main.go
```diff
@@ -209,6 +256,94 @@ func makeVulnerability(initems []gozdef.VulnEvent, cmd mig.Command) (items []goz
...
+	return
+}
+
+// query a ServiceAPI instance for the set of all assets
+// load them into a searchable map, keyed to asset hostname
+// the ServiceAPI object must already be loaded with a Bearer token
+func GetAssets(m map[string]ServiceApiAsset, api ServiceApi) (err error){
+	
+	// get json array of assets from serviceapi
+	requestURL := api.URL + "api/v1/assets/"
```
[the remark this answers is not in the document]
tristanweir: I added some code to cast it as a net/url.URL here (as opposed to in the config object itself). It's not ideal but it should do for now.

runner-plugins/runner-scribe/main.go
tristanweir: I worked on this for a bit, but there is something weird about how ServiceAPI is presenting back the JSON, requiring double-unmarshaling. I couldn't get it to come in with decoding. 

I'd like to defer this change, since we control both ends of the information. Maybe a fix on the serviceAPI side would solve it easier.

runner-plugins/runner-scribe/main.go
```diff
@@ -33,6 +35,34 @@ type config struct {
 		URL      string // URL to post events to MozDef
 		UseProxy bool   // A switch to enable/disable the use of a system-configured proxy
 	}
+	api 		ServiceApi
+}
+
+type ServiceApiAsset struct {
+	Id 				string
```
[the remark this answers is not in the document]
tristanweir: TIL about gofmt. Fixed

runner-plugins/runner-scribe/main.go
```diff
@@ -33,6 +35,34 @@ type config struct {
...
+	Asset_type 		string
+	Asset_identifier string
```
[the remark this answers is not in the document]
tristanweir: Done

runner-plugins/runner-scribe/main.go
```diff
@@ -33,6 +35,34 @@ type config struct {
...
+type ServiceApi struct {
+	URL				string
+	AuthEndpoint 	string
+	ClientID 		string
+	ClientSecret	string
+	Token 			string // ephemeral token we generate to connect to ServiceAPI
+}
+
+type Auth0Token struct {
+	Access_token	string
```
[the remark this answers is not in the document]
tristanweir: Done

tristanweir:
@arcrose Added tests for LookupOperatorTeam. I'm open to suggestions on how to add tests when we call to a specific external API, like we do in GetAuthToken() or GetAssets()

PR #486, opened 2019-01-02
Clean up SystemD agents for smooth re-installs

arcrose:
This PR is a work in progress. Please do not review or merge it until the WIP status is revoked.

At Mozilla, we have encountered an issue with agents in production wherein updates fail due to hanging `systemctl stop` processes.  In order to be able to cleanly reinstall the agent on a host in such circumstances, we need to kill these hanging processes. We also opt to remove the `/etc/mig` directory and its contents as well as the agent's Unit file.  By doing so, automated processes using tools such as Ansible can more-or-less blindly recreate these resources using the latest versions of them, facilitating a smooth re-install.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -184,19 +184,19 @@ rpm-agent: mig-agent
 	make agent-install-script-linux
 	make agent-remove-script-linux
 	fpm \
-    -C tmp \
-    -n mig-agent \
-    --license GPL \
-    --vendor mozilla \
-    --description "Mozilla InvestiGator Agent" \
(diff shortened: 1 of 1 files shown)


ameihm0912:
@arcrose curious on this one, do we know what the root cause of the hanging systemctl process is?

Is it a hang on actual shutdown of the agent or possibly a bad return code when it should be a normal exit?

arcrose:
I'm honestly not 100% sure at this point. I haven't had a chance to fully explore what the cause is,  but there seem to be a couple of problems related to agent updates and the replacement of the Unit file.

PR #487, opened 2019-01-04
Agent api proxying

arcrose:
In this PR, I've moved a little code around with the goal of creating some reusable code for having the agent talk to the MIG API.  I also re-worked the heartbeat code in the agent to use the new `api.Client` abstraction, which means that `POST` requests to upload a heartbeat will now be proxied.  I have tested this out manually on an internal host and verified that:

* An older version of the agent panics upon trying to send a heartbeat
* The new agent does not panic, and continues to operate as normal

### DIFF
--- a/mig-agent/api/api.go
+++ b/mig-agent/api/api.go
@@ -1,0 +1,148 @@
+package api
+
+import (
+	"bytes"
+	"encoding/json"
+	"errors"
+	"fmt"
+	"io"
+	"net"
+	"net/http"
+	"net/url"
+	"strings"
+	"time"
+)
+
+// Each endpoint format string assumes that the base address for the API will
+// be something like "https://domain.com/api/v1".
+const (
+	heartbeatEndptFmt string = "%s/heartbeat"
+)
+
+// Client is an HTTP client abstraction for the MIG API.
+// It must be configured with:
+//
+// 1. The base address of the API, such as https://api.mig.mozilla.com/api/v1
+// 2. A list of addresses of proxies to try to use.
+//
+// Note that the base address MUST NOT contain a trailing forward-slash (/).
+// Note also that the client will try each proxy in order.
+type Client struct {
+	baseAddress    string
+	proxyAddresses []string
+}
+
+// Heartbeat contains information describing an active agent.
+type Heartbeat struct {
+	Name        string      `json:"name"`
+	Mode        string      `json:"mode"`
+	Version     string      `json:"version"`
+	PID         uint        `json:"pid"`
+	QueueLoc    string      `json:"queueLoc"`
+	StartTime   time.Time   `json:"startTime"`
+	RefreshTime time.Time   `json:"refreshTime"`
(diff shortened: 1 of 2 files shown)


mig-agent/api/api.go
```diff
@@ -0,0 +1,148 @@
...
+	method string,
+	destUrl string,
+	body io.Reader,
+	response interface{},
+) (int, error) {
+	var httpClient *http.Client = nil
+	var transport http.RoundTripper = http.DefaultTransport
+
+	var didFindProxy = len(client.proxyAddresses) == 0
+	for _, proxyAddr := range client.proxyAddresses {
```
tristanweir: The logic here is a little confusing. As I read it, we check to make sure that there is a parsable proxy URL in the set described by configuration, and we load the first one. 

However the check on !didFindProxy immediately following this block indicates that "None of the configured proxies was found to be usable." It doesn't appear we validate the proxy is reachable, just that it isn't non-URL input.

Maybe some additional commenting or modification of the error message would be helpful.

mig-agent/api/api.go
```diff
@@ -0,0 +1,148 @@
...
+	}
+}
+
+// Heartbeat posts a heartbeat message to the MIG API, indicating that the
+// agent is active.
+func (client Client) Heartbeat(hb Heartbeat) error {
+	fullURL := fmt.Sprintf(heartbeatEndptFmt, client.baseAddress)
+	encoded, _ := json.Marshal(hb)
+	body := bytes.NewReader(encoded)
+	response := struct {
```
tristanweir: In this case, I believe response will only get loaded with the error message, not the Response Body from the HTTP POST. Is that intended? Would changing 'response' to 'responseErr' clarify that? Since sendJSON is generalized, should we allow for returning the full response (status, error, body, etc.)

mig-agent/api/api.go
```diff
@@ -0,0 +1,148 @@
+package api
```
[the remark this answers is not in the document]
arcrose: I think it's fine in this case because it's scoped under `mig-agent/api`. There shouldn't be any confusion between `mig/api` and `mig/mig-agent/api` but if there is, one can always rename the import by writing
```go
import (
  apiInterface "github.com/mozilla/mig/mig-agent/api"
)

// ...

apiInterface.NewClient(...) // instead of api.NewClient
```

mig-agent/api/api.go
arcrose: > It doesn't appear we validate the proxy is reachable, just that it isn't non-URL input.

This is a good point. That's my mistake.

mig-agent/api/api.go
arcrose: `sendJSON` is actually JSON-deserializing the body of the response it gets into whatever container you pass to it. So in this case, since `response` is a struct with a field tagged for JSON (de)serialization, it will expect to deserialize the body as a JSON object containing an `error` field of type `string | null`.

PR #489, opened 2019-01-08
How do I keep missing these dashes

arcrose:


Makefile
```diff
@@ -188,13 +188,13 @@ rpm-agent: mig-agent
 		-n mig-agent \
 		--license GPL \
 		--vendor mozilla \
-		---description "Mozilla InvestiGator Agent" \
+		--description "Mozilla InvestiGator Agent" \
```
floatingatoll: To note something perhaps unconsidered, this static string is repeated many times in this Makefile in various build sections. Consider trying to reduce this over time using make variables of some sort.

